PR #6, opened 2021-06-07
feat(replay): implement replay dry-run feature

mauliksoneji:


models/dag_node.go
```diff
@@ -0,0 +1,75 @@
+package models
```
kushsharma, line 1: this is not the right place to put this I think, models only contain business models, DAGNode doesn't look like one

kushsharma (changes requested):

api/handler/v1/runtime.go
```diff
@@ -729,6 +729,50 @@ func (sv *RuntimeServiceServer) ListResourceSpecification(ctx context.Context, r
...
+	if req.EndDate != "" {
+		if endDate, err = time.Parse(job.DateFormat, req.EndDate); err != nil {
+			return nil, fmt.Errorf("unable to parse replay end date(e.g. %s): %v", job.DateFormat, err)
+		}
+	}
+	if endDate.Before(startDate) {
+		return nil, errors.New("replay end date cannot be before start date")
+	}
+
+	rootNode, err := sv.jobSvc.Replay(namespaceSpec, jobSpec, startDate, endDate)
```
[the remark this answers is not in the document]
mauliksoneji: this is resolved, please check

api/handler/v1/runtime.go
```diff
@@ -729,6 +729,50 @@ func (sv *RuntimeServiceServer) ListResourceSpecification(ctx context.Context, r
...
+	}
+
+	endDate := startDate
+	if req.EndDate != "" {
+		if endDate, err = time.Parse(job.DateFormat, req.EndDate); err != nil {
+			return nil, fmt.Errorf("unable to parse replay end date(e.g. %s): %v", job.DateFormat, err)
+		}
+	}
+	if endDate.Before(startDate) {
+		return nil, errors.New("replay end date cannot be before start date")
```
kushsharma, line 763: use `status` package here instead of returning `error`

api/handler/v1/runtime.go
```diff
@@ -729,6 +729,50 @@ func (sv *RuntimeServiceServer) ListResourceSpecification(ctx context.Context, r
...
+	}
+
+	rootNode, err := sv.jobSvc.Replay(namespaceSpec, jobSpec, req.DryRun, startDate, endDate)
+	if err != nil {
+		return nil, fmt.Errorf("error while processing replay: %v", err)
+	}
+	return &pb.ReplayResponse{
+		Success:  true,
+		Response: rootNode.ToReplayResponseNode(),
```
kushsharma, line 772: `ToReplayResponseNode` attaching this in the main model, leaks protobuf implementation details to internal packages. So far we have been using adapters to convert them, this keeps optimus models decoupled with proto models. I think we should extract this as a function instead of adding it as part of the struct.

api/handler/v1/runtime.go
```diff
@@ -729,6 +729,50 @@ func (sv *RuntimeServiceServer) ListResourceSpecification(ctx context.Context, r
...
+			return nil, fmt.Errorf("unable to parse replay end date(e.g. %s): %v", job.DateFormat, err)
+		}
+	}
+	if endDate.Before(startDate) {
+		return nil, errors.New("replay end date cannot be before start date")
+	}
+
+	rootNode, err := sv.jobSvc.Replay(namespaceSpec, jobSpec, req.DryRun, startDate, endDate)
+	if err != nil {
+		return nil, fmt.Errorf("error while processing replay: %v", err)
```
kushsharma, line 768: `status` package, no need to say its an `error` in error strings, just mention what the error is or when it happened.

api/proto/odpf/metadata/optimus/Job.pb.go
```diff
@@ -1,18 +1,18 @@
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
-// 	protoc-gen-go v1.25.0
-// 	protoc        v3.13.0
+// 	protoc-gen-go v1.26.0-rc.1
```
kushsharma, line 3: can you try running `make install` then generate protos? I think I have pinned the compiler version, not sure why it's generated with a different one.

cmd/replay.go
```diff
@@ -0,0 +1,208 @@
...
+	if _, ok := taskReruns[instance.JobName]; !ok {
+		taskReruns[instance.JobName] = taskRunBlock{
+			name:   instance.JobName,
+			height: height,
+			runs:   set.NewTreeSetWithTimeComparator(),
+		}
+	}
+
+	for _, taskRun := range instance.Runs {
+		parsedTime, _ := time.Parse(time.RFC3339, taskRun)
```
kushsharma, line 50: if taskRun is just a time.RFC3339 timestamp, why not use timestamp protobuf for this? Why string?

cmd/replay.go
```diff
@@ -0,0 +1,208 @@
...
+			return nil
+		},
+	}
+	reCmd.Flags().BoolVarP(&dryRun, "dry-run", "", dryRun, "do a trial run with no permanent changes")
+	reCmd.Flags().StringVarP(&replayProject, "project", "p", "", "project name of optimus managed ocean repository")
+	reCmd.MarkFlagRequired("project")
+	reCmd.Flags().StringVarP(&namespace, "namespace", "n", "", "namespace of deployee")
+	reCmd.MarkFlagRequired("namespace")
+
+	reCmd.Run = func(cmd *cli.Command, args []string) {
```
kushsharma, line 101: Use `RunE` and return error

cmd/replay.go
```diff
@@ -0,0 +1,208 @@
...
+		endDate := args[1]
+		if len(args) >= 3 {
+			endDate = args[2]
+		}
+		if err := replayJobBuildRequest(l, replayProject, namespace, args[0], args[1], endDate, dryRun, conf); err != nil {
+			l.Println(err)
+			os.Exit(1)
```
kushsharma, line 108: No need to print error, no need to exit, just return error.

job/replay.go
```diff
@@ -0,0 +1,187 @@
...
+
+	"github.com/odpf/optimus/core/cron"
+	"github.com/odpf/optimus/core/multi_root_tree"
+	"github.com/odpf/optimus/models"
+	"github.com/pkg/errors"
+)
+
+const (
+	// DateFormat YYYY-mm-dd for replay dates and dag start date
+	DateFormat = "2006-01-02"
```
kushsharma, line 15: Can we prefix it with `Replay` so that it's clear it's for replay only?

models/job.go
```diff
@@ -279,6 +279,8 @@ type JobService interface {
 	GetByName(string, NamespaceSpec) (JobSpec, error)
 	// Dump returns the compiled Job
 	Dump(NamespaceSpec, JobSpec) (Job, error)
+	// Replay replays the jobSpec and its dependencies between start and endDate
+	Replay(NamespaceSpec, JobSpec, bool, time.Time, time.Time) (*DAGNode, error)
```
kushsharma, line 283: I am not sure if the response of this interface method that is `DAGNode` fits here. I mean, replay is a lot more than just the execution tree.

models/dag_node.go
```diff
@@ -0,0 +1,47 @@
...
+import (
+	"time"
+
+	pb "github.com/odpf/optimus/api/proto/odpf/optimus"
+	"github.com/odpf/optimus/core/set"
+)
+
+// DAGNode represents a custom data type that contains a DAGSpec along with it's dependent DAGNodes
+type DAGNode struct {
+	DAG        JobSpec
```
kushsharma, line 12: What if we make it generic and use an interface here? Just a suggestion, not sure. This can be moved to core then.

job/replay.go
```diff
@@ -0,0 +1,187 @@
...
+func populateDownstreamDAGs(tree *multi_root_tree.MultiRootDAGTree, jobSpec models.JobSpec, dagSpecMap map[string]models.JobSpec) (*models.DAGNode, error) {
+	for _, childSpec := range dagSpecMap { // (1) -> 2 -> 3
+		childNode := findOrCreateDAGNode(tree, childSpec) // 3
+		for _, depDAG := range childSpec.Dependencies {
+			var isExternal = false
+			parentSpec, ok := dagSpecMap[depDAG.Job.Name] // 2
+			if !ok {
+				if depDAG.Type == models.JobSpecDependencyTypeIntra {
+					return nil, errors.Wrap(ErrJobSpecNotFound, depDAG.Job.Name)
+				} else {
```
kushsharma, line 92: we can pull the else statement out if we are returning in the base if statement.

kushsharma (changes requested):

api/handler/v1/runtime.go
mauliksoneji: fixed

api/handler/v1/runtime.go
mauliksoneji: fixed

models/job.go
mauliksoneji: updated

job/replay.go
mauliksoneji: updated

api/handler/v1/runtime.go
mauliksoneji: updated

cmd/replay.go
mauliksoneji: updated

job/replay.go
mauliksoneji: yes, removed

cmd/replay.go
mauliksoneji: updated

models/dag_node.go
mauliksoneji: I have moved the dag_node file to core package

models/dag_node.go
mauliksoneji: This is a good idea, I have made this change

cmd/replay.go
mauliksoneji: I have made this change

models/job.go
mauliksoneji: for now since we just have dry run functionality, lets just send the execution tree. Later we will add other values

core/multi_root_tree/multi_root_dag_tree.go
```diff
@@ -0,0 +1,101 @@
...
+
+var (
+	// ErrCyclicDependencyEncountered is triggered a tree has a cyclic dependency
+	ErrCyclicDependencyEncountered = errors.New("a cycle dependency encountered in the tree")
+)
+
+// MultiRootDAGTree - represents a data type which has multiple independent root nodes
+// all root nodes have their independent tree based on depdencies of TreeNode.
+// it also maintains a map of nodes for faster lookups and managing node data.
+type MultiRootDAGTree struct {
```
kushsharma, line 15: Struct name and package name are prefixed with the same literal and will stutter when used. Can we rename it to something else?

mauliksoneji: we can call this package as tree

models/job.go
kushsharma: What do you have in mind? So are we going to have two functions like `ReplayDryRun` and `Replay` or just `Replay` with a composite response that will have a field `DryRun` true along with a execution tree? The thing is this is also indirectly affecting your `proto` for replay, and even though its easy to do changes in `interface`, we will have to do breaking changes in proto as well.

core/multi_root_tree/multi_root_dag_tree.go
mauliksoneji: updated

models/job.go
mauliksoneji: made this change

mauliksoneji:
Updated changelog from previous review:
- Replace ReplayRequest as ReplayDryRunRequest
- Made generic dagNode and multiRootTree to accept any data item
- Send start and end time as protobuf.Timestamp instead of string

kushsharma (commented):
LGTM

PR #22, opened 2021-07-03
test: add test cases of adapter in hook

scortier:
Added the remaining edge case in adapter file and the test cases of all the functions in adapter file.

### DIFF
--- a/plugin/cli/adapter_test.go
+++ b/plugin/cli/adapter_test.go
@@ -1,0 +1,510 @@
+package cli
+
+import (
+	"reflect"
+	"testing"
+
+	pb "github.com/odpf/optimus/api/proto/odpf/optimus/plugins"
+	"github.com/odpf/optimus/models"
+)
+
+func Test_AdaptQuestionToProto(t *testing.T) {
+	type args struct {
+		conf models.PluginQuestion
+	}
+	tests := []struct {
+		name string
+		args args
+		want *pb.PluginQuestion
+	}{
+		{
+			name: "should work for non empty inputs (with empty SubQuestion parameter)",
+			args: args{
+				conf: models.PluginQuestion{
+					Name:         "key",
+					Prompt:       "sec",
+					Help:         "do",
+					Default:      "this",
+					Multiselect:  nil,
+					SubQuestions: nil,
+				},
+			},
+			want: &pb.PluginQuestion{
+				Name:         "key",
+				Prompt:       "sec",
+				Help:         "do",
+				Default:      "this",
+				Multiselect:  nil,
+				SubQuestions: nil,
+			},
+		},
+		{
+			name: "should work for non empty inputs",
+			args: args{
+				conf: models.PluginQuestion{
+					Name:        "key",
+					Prompt:      "sec",
+					Help:        "do",
+					Default:     "this",
+					Multiselect: nil,
+					SubQuestions: []models.PluginSubQuestion{
+						{
+							IfValue:   "",
+							Questions: []models.PluginQuestion{},
+						},
+					},
+				},
+			},
+			want: &pb.PluginQuestion{
+				Name:        "key",
+				Prompt:      "sec",
+				Help:        "do",
+				Default:     "this",
+				Multiselect: nil,
+				SubQuestions: []*pb.PluginQuestion_SubQuestion{
+					{
+						IfValue:   "",
+						Questions: []*pb.PluginQuestion{},
+					},
+				},
+			},
+		},
+		{
+			name: "should work for non empty inputs",
+			args: args{
+				conf: models.PluginQuestion{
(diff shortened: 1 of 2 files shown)


plugin/hook/adapter_test.go
```diff
@@ -0,0 +1,622 @@
...
+							Multiselect:  nil,
+							SubQuestions: nil,
+						},
+					},
+				},
+			},
+		}
+		assert.Equal(t, expected, actual)
+	})
+	t.Run("should work for non empty inputs", func(t *testing.T) {
```
kushsharma, line 113: whats the difference between this and line 66 test

plugin/hook/adapter_test.go
```diff
@@ -0,0 +1,622 @@
...
+			Name:         "key",
+			Prompt:       "sec",
+			Help:         "this",
+			Default:      "do",
+			Multiselect:  nil,
+			SubQuestions: []models.PluginSubQuestion{},
+		}
+		assert.Equal(t, expected, actual)
+	})
+	t.Run("should work for nil input", func(t *testing.T) {
```
kushsharma, line 187: This test and the test at 168 is same. In Go, `"nil"` is not equal to `nil`

plugin/hook/adapter_test.go
```diff
@@ -0,0 +1,622 @@
...
+					Value: "",
+				},
+			},
+		}
+		assert.Equal(t, expected, actual)
+	})
+	t.Run("should work for non-empty input", func(t *testing.T) {
+		actual := adaptConfigsToProto(models.HookPluginConfigs{
+			models.HookPluginConfig{
+				Name:  "nil",
```
kushsharma, line 344: Remove all usages of `"nil"` that string containing `nil` word as it is confusing.

plugin/hook/adapter_test.go
```diff
@@ -0,0 +1,622 @@
...
+			},
+		})
+		expected := models.HookPluginConfigs{
+			models.HookPluginConfig{
+				Name: "key",
+			},
+		}
+		assert.Equal(t, expected, actual)
+	})
+	t.Run("partial config should be adapted from proto correctly", func(t *testing.T) {
```
kushsharma, line 452: this test and the test above this can be merged into one, there are a lot of test cases that can be merged into one instead of writing them separately. Kindly merge them wherever possible.

kushsharma (changes requested):

plugin/hook/adapter_test.go
scortier: On line 66 I have passes nil to the Question object while in 133 it is any random para (non-zero)

plugin/hook/adapter.go
```diff
@@ -55,6 +58,11 @@ func adaptQuestionFromProto(q *pb.PluginQuestion) models.PluginQuestion {
 }
 
 func adaptConfigsToProto(c models.HookPluginConfigs) *pb.HookConfigs {
+	if c == nil {
```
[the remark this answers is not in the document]
scortier: okay

plugin/hook/adapter.go
```diff
@@ -79,6 +90,11 @@ func adaptConfigFromProto(a *pb.HookConfigs) models.HookPluginConfigs {
 }
 
 func adaptAssetsToProto(a models.HookPluginAssets) *pb.HookAssets {
+	if a == nil {
```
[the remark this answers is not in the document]
scortier: got it

plugin/hook/adapter_test.go
scortier: Is `SubQuestions: nil,` can be written as `SubQuestions: []*pb.PluginQuestion_SubQuestion{},` ?

plugin/hook/adapter_test.go
scortier: done

plugin/hook/adapter_test.go
scortier: Is checking only 1 parameter here will cover the case ?

plugin/hook/server_test.go
```diff
@@ -0,0 +1,181 @@
...
+		request := &pb.GetHookQuestions_Request{}
+		response, err := server.GetHookQuestions(ctx, request)
+		assert.Nil(t, err)
+		assert.Equal(t, hookQuestionResponse.Questions, response.Questions)
+
+	})
+
+}
+
+func TestGRPCServer_ValidateHookQuestion(t *testing.T) {
```
scortier, line 97: @mauliksoneji  can you help me with `func TestGRPCServer_ValidateHookQuestion` not able to identify what is the error (tried some changes in request (line 114) but not able to figure out). 
Also not able to figure out the how to check args (from line 103 )

PR #35, opened 2021-07-13
feat: validate replay request before execution

mauliksoneji:

PR #37, opened 2021-07-16
refactor: map replay errors, add replay failure test cases, refactor airflow and replay manager

arinda-arif:
based on additional feedback in https://github.com/odpf/optimus/pull/35


api/handler/v1/runtime.go
```diff
@@ -794,6 +794,11 @@ func (sv *RuntimeServiceServer) Replay(ctx context.Context, req *pb.ReplayReques
 
 	replayUUID, err := sv.jobSvc.Replay(ctx, replayWorkerRequest)
 	if err != nil {
+		if err == job.ErrRequestQueueFull {
```
kushsharma, line 797: I think we should use `errors.Is()` instead of directly comparing them with `==` so that if the upstream function wraps them in another error, it should still work

kushsharma (changes requested):

kushsharma:
Can you also change the commit message and PR message to actually reflect what we are doing in this PR. Also, this looks like more of a refactor than fix.

arinda-arif:
Hello @kushsharma! Have made the changes as you requested, can you please check? Thanks!

kushsharma (approved):

Issue #44, opened 2021-07-25
Implement google sheets external data table type for bigquery datastore


vianhazman:
Enable google sheets externaltables management for BigQuery datastore via Optimus.

User should be able to:
- Create a google sheets external table by specifying sheets URI
- Define schema for the google sheets external table
- Use metadata management feature such as Labels

The implementation should be able to extend other BigQuery supported external data sources for future development.
About BigQuery external tables: https://cloud.google.com/bigquery/docs/external-tables

kushsharma:
@vianhazman this looks like a nice feature to have and I can see you have already opened a PR for it. Can you also provide a sample of YAML specification that it will use?

vianhazman:
Sure, the following specification is specific for Google Sheets source. In the future, we can add more source types such as Google Drive and GCS with a different configuration in the `config` property:

```
name: project.dataset.table
type: external_table
version: 1
spec:
  schema:
  - name: column_name
    type: STRING
  source: 
    type: google_sheets
    uris:
    - https://docs.google.com/spreadsheets/d/spreadsheet_id
    config:
      range: Sheet1!A1:B4
      skip_leading_rows: 1
```

PR #45, opened 2021-07-25
feat: google sheets external table support for bigquery datastore

vianhazman:
Resolve https://github.com/odpf/optimus/issues/44


ext/datastore/bigquery/table_spec.go
```diff
@@ -145,7 +148,7 @@ func (s tableSpecHandler) ToYaml(optResource models.ResourceSpec) ([]byte, error
 func (s tableSpecHandler) FromYaml(b []byte) (models.ResourceSpec, error) {
 	var yamlResource TableResourceSpec
 	if err := yaml.Unmarshal(b, &yamlResource); err != nil {
-		return models.ResourceSpec{}, err
+		return models.ResourceSpec{}, nil
```
kushsharma, line 151: Is this done by mistake?

vianhazman: Oops yes it was done by mistake, thanks for pointing out

ext/datastore/bigquery/adapter.go
```diff
@@ -84,6 +84,73 @@ func bqFieldModeTo(field BQField) (fieldMode, error) {
...
+		Range:           sheetRange,
+	}
+}
+
+func bqGoogleSheetsOptionsFrom(opt *bqapi.GoogleSheetsOptions) map[string]interface{} {
+	resultMap := make(map[string]interface{})
+
+	if opt.SkipLeadingRows != 0 {
+		// Map value of int has to be converted to float because of using interface{}
+		resultMap["skip_leading_rows"] = float64(opt.SkipLeadingRows)
```
kushsharma, line 110: Seeing this for the first time, what's the problem with having `interface{}` if we don't convert it into `float64`?

ext/datastore/bigquery/adapter.go
```diff
@@ -84,6 +84,73 @@ func bqFieldModeTo(field BQField) (fieldMode, error) {
...
+		Options:      option,
+	}
+	return externalConfig, nil
+}
+
+func bqExternalDataConfigFrom(c *bqapi.ExternalDataConfig) (*BQExternalSource, error) {
+	var option map[string]interface{}
+
+	switch ExternalTableType(c.SourceFormat) {
+	case ExternalTableTypeGoogleSheets:
```
kushsharma, line 141: Can't we use `bqapi.GoogleSheets` here as well like line number 122

ext/datastore/bigquery/adapter.go
```diff
@@ -159,6 +226,14 @@ func bqCreateTableMetaAdapter(t BQTable) (meta *bqapi.TableMetadata, err error)
 			meta.RangePartitioning = bqPartitioningRangeTo(*t.Metadata.Partition)
 		}
 	}
+
+	if t.Metadata.Source != nil {
+		meta.ExternalDataConfig, err = bqExternalDataConfigTo(*t.Metadata.Source)
```
kushsharma, line 231: So how is the table schema gets populated in this case? Is the user going to pass schema just like a normal table?

ext/datastore/bigquery/adapter_test.go
```diff
@@ -67,6 +67,34 @@ func TestAdapter(t *testing.T) {
...
+			SourceFormat: bigquery.GoogleSheets,
+			SourceURIs:   []string{"http://googlesheets.com/1234"},
+			Options: &bigquery.GoogleSheetsOptions{
+				SkipLeadingRows: 1,
+				Range:           "A!:A1:B1",
+			},
+		}
+
+		bQExternalDataConfigResult, err := bqExternalDataConfigTo(externalDataSource)
+
```
kushsharma, line 86: can you please format to remove these empty new lines in between?

ext/datastore/bigquery/external_table_spec.go
```diff
@@ -0,0 +1,71 @@
...
+	return map[string]string{}
+}
+
+func extractTableSourceFromProtoStruct(protoVal *structpb.Value) *BQExternalSource {
+	sInfo := &BQExternalSource{}
+	if protoVal.GetStructValue() == nil {
+		return sInfo
+	}
+	if f, ok := protoVal.GetStructValue().Fields["type"]; ok {
+		sInfo.SourceType = f.GetStringValue()
```
kushsharma, line 56: don't we need to uppercase the value to match with our type in case it's not already?

ext/datastore/bigquery/adapter.go
vianhazman: Yes, there are two ways on defining schema for external tables:
- Use autodetect schema It will disable the schema provided  and BQ will attempt to detect the schema by the sheets
- Use user defined schema, BQ will map the google sheets column to the defined schema
On this PR, we detect schema based on user passed schema. For the next development. we can implement the autodetect schema by having the `AutoDetect` option on the `ExternalDataConfig` as true if no schema is supplied by the user.

ext/datastore/bigquery/adapter_test.go
vianhazman: Will refactor this

ext/datastore/bigquery/adapter.go
vianhazman: Agree, will refactor this

ext/datastore/bigquery/external_table_spec.go
vianhazman: Agreed that such validation should be there. Will add

ext/datastore/bigquery/adapter.go
vianhazman: The `interface{}` type always asserts numbers from marshalled `[]byte` as `float64`. Have tried before to directly using `int64` or `int` but it seems both throws an error.
```
panic: interface conversion: interface {} is float64, not int
```
```
panic: interface conversion: interface {} is float64, not int64
```
Do you have any suggestion on this?

kushsharma: Got it. Yeah JSON unmarshal will default it to float64, it's fine I guess when we unmarshal(line number 92) but when we are creating this map, we don't need to cast it to float64.

ext/datastore/bigquery/adapter_test.go
```diff
@@ -67,6 +67,29 @@ func TestAdapter(t *testing.T) {
 		},
 	}
 
+	t.Run("should convert from and to BQ ExternalDataConfig successfully", func(t *testing.T) {
+		externalDataSource := BQExternalSource{
+			SourceType: string(ExternalTableTypeGoogleSheets),
+			SourceURIs: []string{"http://googlesheets.com/1234"},
+			Config:     map[string]interface{}{"skip_leading_rows": 1.0, "range": "A!:A1:B1"},
```
kushsharma, line 74: here as well, just set the row count as `1`

Issue #27, opened 2021-07-07
Optimus CLI should check for updates using Github releases


kushsharma:
Having a notification about an update available for optimus cli can be helpful for users to keep the binary up to date. We can do this using github release api I think. This notification can be shown to users maybe once a week.

scortier:
PR #46

scortier:
This issue can be closed now as the PR has been merged.

PR #46, opened 2021-07-28
feat: support latest binary alert feature through cli 

scortier:
### Description
If the user wants to check whether any latest binary release is available (it can be minor release too).

**Just need to trigger the command to check the released update : `optimus version`**

### Preview

Screenshot 2021-07-28 at 10 16 19 AM

### Related Issues 
Issue #27

### DIFF
--- a/cmd/version.go
+++ b/cmd/version.go
@@ -2,12 +2,16 @@ package cmd
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
+	"io/ioutil"
+	"net/http"
 	"time"
 
-	"github.com/odpf/optimus/models"
+	"github.com/hashicorp/go-version"
 
 	"github.com/odpf/optimus/config"
+	"github.com/odpf/optimus/models"
 
 	pb "github.com/odpf/optimus/api/proto/odpf/optimus"
 	"github.com/pkg/errors"
@@ -16,7 +20,9 @@ import (
 )
 
 const (
-	versionTimeout = time.Second * 2
+	versionTimeout   = time.Second * 2
+	clientVersion    = time.Second * 1
+	githubReleaseURL = "https://api.github.com/repos/odpf/optimus/releases/latest"
 )
 
 func versionCommand(l logger, host string, pluginRepo models.PluginRepository) *cli.Command {
@@ -26,7 +32,6 @@ func versionCommand(l logger, host string, pluginRepo models.PluginRepository) *
 		Short: "Print the client version information",
 		RunE: func(c *cli.Command, args []string) error {
 			l.Printf(fmt.Sprintf("client: %s-%s", coloredNotice(config.Version), config.BuildCommit))
-
 			if host != "" && serverVersion {
 				srvVer, err := getVersionRequest(config.Version, host)
 				if err != nil {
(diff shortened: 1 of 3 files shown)


scortier:
Please review the PR @kushsharma

kushsharma (changes requested):

cmd/version.go
```diff
@@ -1,10 +1,18 @@
+//version.go
```
kushsharma, line 1: Why is this here?

cmd/version.go
```diff
@@ -15,8 +21,45 @@ import (
...
 )
 
+func checkLatestVersion() string {
+	gitClient := http.Client{
+		Timeout: clientVersion,
+	}
+
+	req, err := http.NewRequest(http.MethodGet, apiURL, nil)
+	if err != nil {
+		log.Warn("failed to create request for latest version")
```
kushsharma, line 35: If an error occurs it should not continue to execute the function and return an empty string

cmd/version.go
```diff
@@ -15,8 +21,45 @@ import (
 
 const (
 	versionTimeout = time.Second * 2
+	clientVersion  = time.Second * 1
+	apiURL         = "https://api.github.com/repos/odpf/optimus/releases/latest"
 )
 
+func checkLatestVersion() string {
```
kushsharma, line 28: lets move this function after `func versionCommand`

cmd/version.go
```diff
@@ -25,6 +68,18 @@ func versionCommand(l logger, host string) *cli.Command {
...
+			currentV, err := version.NewVersion(config.Version)
+			if err != nil {
+				return errors.Wrapf(err, "failed to parse current version")
+			}
+			latestV, err := version.NewVersion(checkLatestVersion())
+			if err != nil {
+				return errors.Wrapf(err, "failed to parse latest version")
+			}
+			if currentV.LessThan(latestV) {
+				l.Printf(fmt.Sprintf("new version is available : %s , consider updating the client", coloredNotice(latestV)))
```
kushsharma, line 80: why space before comma?

cmd/version.go
```diff
@@ -25,6 +68,18 @@ func versionCommand(l logger, host string) *cli.Command {
 		RunE: func(c *cli.Command, args []string) error {
 			l.Printf(fmt.Sprintf("client: %s-%s", coloredNotice(config.Version), config.BuildCommit))
 
+			currentV, err := version.NewVersion(config.Version)
+			if err != nil {
+				return errors.Wrapf(err, "failed to parse current version")
+			}
+			latestV, err := version.NewVersion(checkLatestVersion())
+			if err != nil {
+				return errors.Wrapf(err, "failed to parse latest version")
```
kushsharma, line 77: when `checkLatestVersion` will return an empty string at the time of error, we should not return error here, simply ignore. We want this check for update flow non-obstructive

go.mod
```diff
@@ -17,12 +15,14 @@ require (
 	github.com/golang/protobuf v1.5.2
 	github.com/google/uuid v1.1.2
 	github.com/googleapis/google-cloud-go-testing v0.0.0-20200911160855-bcd43fbb19e8
+	github.com/gorilla/mux v1.7.4
```
kushsharma, line 18: I don't think we use mux anywhere, can you consider running `go mod tidy`?

kushsharma (changes requested):

cmd/version.go
scortier: But i think we discussed, that the execution of the command should not stop regardless of the error , we will just warn the user about the error

kushsharma: we are not stopping the execution of the command but how do you expect this function to work if it failed to create even a valid http request.

cmd/version.go
```diff
@@ -27,6 +34,18 @@ func versionCommand(l logger, host string, pluginRepo models.PluginRepository) *
...
+			}
+
+			if currentV.LessThan(latestV) {
+				l.Printf(fmt.Sprintf("new version is available : %s, consider updating the client", coloredNotice(latestV)))
+			}
```
kushsharma, line 48: what if we move all this to `checkLatestVersion` and take `config.Version` as an argument? This will avoid failing for errors when parsing the latest version at line number 43.

kushsharma (changes requested):

PR #67, opened 2021-08-23
feat: add replay list sub command

arinda-arif:
Add replay list sub command to get list of recent replays per project.

**Usage:**
optimus replay list --project [project_name]

### DIFF
--- a/api/handler/v1/runtime.go
+++ b/api/handler/v1/runtime.go
@@ -801,7 +801,7 @@ func (sv *RuntimeServiceServer) GetReplayStatus(ctx context.Context, req *pb.Get
 		return nil, err
 	}
 
-	replayState, err := sv.jobSvc.GetStatus(ctx, replayRequest)
+	replayState, err := sv.jobSvc.GetReplayStatus(ctx, replayRequest)
 	if err != nil {
 		return nil, status.Errorf(codes.Internal, "error while getting replay: %v", err)
 	}
@@ -841,7 +841,7 @@ func (sv *RuntimeServiceServer) ListReplays(ctx context.Context, req *pb.ListRep
 		return nil, err
 	}
 
-	replays, err := sv.jobSvc.GetList(projSpec.ID)
+	replays, err := sv.jobSvc.GetReplayList(projSpec.ID)
 	if err != nil {
 		return nil, status.Errorf(codes.Internal, "error while getting replay list: %v", err)
 	}
--- a/api/handler/v1/runtime.go
+++ b/api/handler/v1/runtime.go
@@ -754,19 +754,12 @@ func (sv *RuntimeServiceServer) ListResourceSpecification(ctx context.Context, r
 }
 
 func (sv *RuntimeServiceServer) ReplayDryRun(ctx context.Context, req *pb.ReplayDryRunRequest) (*pb.ReplayDryRunResponse, error) {
-	replayWorkerRequest, err := sv.parseReplayRequest(&pb.ReplayRequest{
-		ProjectName: req.ProjectName,
-		JobName:     req.JobName,
-		Namespace:   req.Namespace,
-		StartDate:   req.StartDate,
-		EndDate:     req.EndDate,
-		Force:       false,
-	})
+	replayRequest, err := sv.parseReplayRequest(req.ProjectName, req.Namespace, req.JobName, req.StartDate, req.EndDate, false)
 	if err != nil {
 		return nil, err
 	}
 
-	rootNode, err := sv.jobSvc.ReplayDryRun(replayWorkerRequest)
+	rootNode, err := sv.jobSvc.ReplayDryRun(replayRequest)
 	if err != nil {
 		return nil, status.Errorf(codes.Internal, "error while processing replay dry run: %v", err)
 	}
@@ -782,7 +775,7 @@ func (sv *RuntimeServiceServer) ReplayDryRun(ctx context.Context, req *pb.Replay
 }
 
 func (sv *RuntimeServiceServer) Replay(ctx context.Context, req *pb.ReplayRequest) (*pb.ReplayResponse, error) {
-	replayWorkerRequest, err := sv.parseReplayRequest(req)
+	replayWorkerRequest, err := sv.parseReplayRequest(req.ProjectName, req.Namespace, req.JobName, req.StartDate, req.EndDate, req.Force)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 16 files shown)


store/postgres/replay_repository.go
```diff
@@ -265,3 +265,29 @@ func (repo *replayRepository) GetByProjectIDAndStatus(projectID uuid.UUID, statu
 	}
 	return replaySpecs, nil
 }
+
+func (repo *replayRepository) GetByProjectID(projectID uuid.UUID) ([]models.ReplaySpec, error) {
+	var replays []Replay
+	if err := repo.DB.Preload("Job").Joins("JOIN job ON replay.job_id = job.id").
```
kushsharma, line 271: Do we need to do this explicetly? I think gorm handles this automatically?

store/postgres/replay_repository.go
```diff
@@ -265,3 +265,29 @@ func (repo *replayRepository) GetByProjectIDAndStatus(projectID uuid.UUID, statu
...
+		Where("job.project_id = ? and replay.created_at >= (NOW() - INTERVAL '1 MONTH')", projectID).Order("created_at DESC").Find(&replays).Error; err != nil {
```
kushsharma, line 272: We can either take this duration as argument to this function or just return all, hardcoding this doesn't look good

kushsharma (changes requested):

store/postgres/replay_repository.go
arinda-arif: yes, we need to add the join statement as preload is not actually joining them, thus cannot use `job.project_id` in the where clause ([reference](https://stackoverflow.com/questions/55656002/how-to-select-by-fields-in-preloaded-object))

store/postgres/replay_repository.go
arinda-arif: agree. since the function name is GetByProjectID, I will just return them all and filter for the recent ones in the caller

api/handler/v1/runtime.go
```diff
@@ -846,6 +846,36 @@ func (sv *RuntimeServiceServer) parseReplayStatusRequest(req *pb.GetReplayStatus
 	return replayRequest, nil
 }
 
+func (sv *RuntimeServiceServer) GetReplayList(ctx context.Context, req *pb.GetReplayListRequest) (*pb.GetReplayListResponse, error) {
+	projSpec, err := sv.getProjectSpec(req.ProjectName)
+	if err != nil {
+		return nil, err
+	}
+
+	replayList, err := sv.jobSvc.GetList(projSpec.ID)
```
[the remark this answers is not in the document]
arinda-arif: have renamed this to replays, but renamed the method to `ListReplays`

arinda-arif:
@kushsharma i have pushed all the changes, however we still have unresolved discussion on [proton](https://github.com/odpf/proton/pull/25). I will revert the makefile changes in here after those resolved and merged

Makefile
```diff
@@ -27,7 +27,7 @@ pack-files:
 generate-proto: ## regenerate protos
 	@echo " > cloning protobuf from odpf/proton"
 	@rm -rf proton/
-	@git -c advice.detachedHead=false clone https://github.com/odpf/proton --depth 1 --quiet --branch main
```
ravisuhag, line 30: @arinda-arif Shouldn't this remain main?
@kushsharma

arinda-arif: yes, this will be changed to main once the proto is merged. I can mark this PR to draft to show the pre-requisite is not complete to avoid any confusion.

job/replay_manager.go
```diff
@@ -30,6 +30,8 @@ var (
 	TimestampLogFormat = "2006-01-02T15:04:05+00:00"
 	// schedulerBatchSize number of run instances to be checked per request
 	schedulerBatchSize = 100
+	//replayListWindow window interval to fetch recent replays
+	replayListWindow = -30 * 24 * time.Hour
```
kushsharma, line 34: let's make it 3 months at least as we won't have many replays anyway.

job/replay_manager.go
```diff
@@ -156,6 +158,25 @@ func (m *Manager) GetReplay(replayUUID uuid.UUID) (models.ReplaySpec, error) {
...
+	if err != nil {
+		if err == store.ErrResourceNotFound {
+			return []models.ReplaySpec{}, nil
+		}
+		return []models.ReplaySpec{}, err
+	}
+
+	var recentReplays []models.ReplaySpec
+	for _, replay := range replays {
+		if replay.CreatedAt.After(time.Now().Add(replayListWindow)) {
```
kushsharma, line 173: how are you testing this function if you are using `time.Now()` here?

job/replay_manager.go
arinda-arif: updated

job/replay_manager.go
arinda-arif: testing it this way: https://github.com/odpf/optimus/blob/8c19f89adf3038f7c787a74bdf1650626d584fb2/job/replay_manager_test.go#L374

since this function is not too time sensitive (have wide time window: 90 days), i simply use the current time of the test, and exclude the replay that aged 100 days. let me know if this is not proper.

arinda-arif:
@kushsharma I have rebased, updated the proto commit, and regenerated proto.

job/replay_manager.go
kushsharma: but this test will start to fail after few 3 months because your replay specs are hardcoded with createdAt right? 😄

job/service.go
```diff
@@ -67,6 +67,7 @@ type ReplayManager interface {
 	Init()
 	Replay(context.Context, models.ReplayRequest) (string, error)
 	GetReplay(uuid.UUID) (models.ReplaySpec, error)
+	GetReplayList(uuid.UUID) ([]models.ReplaySpec, error)
```
kushsharma, line 70: whats uuid here?

kushsharma: I think this is project id, on line number 69 because we are taking replay id, its fine to omit the argument name but because here we are taking project id, its not so obvious that is replay id or project id or something else? I think we should add named argument here.

api/handler/v1/runtime.go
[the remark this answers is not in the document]
arinda-arif: also, have renamed:
- `jobSvc.GetList` to `jobSvc.GetReplayList`
- `jobSvc.GetStatus` to `jobSvc.GetReplayStatus`

job/replay_manager.go
arinda-arif: the createdAt replaySpecs in test has value of today minus several days so it wont pass after 3 months 😄 but i'm also fine if the starting window being passed as an argument is better and cleaner

job/service.go
arinda-arif: agree, updated

job/replay_manager.go
kushsharma: oh got it, so even the mocked replays are using `time.Now` on createdAt then I think it's fine. We can leave it like this. 
Also, we should use `UTC` with `Now()`.

arinda-arif: got it. have updated this

kushsharma (approved):

kushsharma:
LGTM

Issue #74, opened 2021-08-30
Allow plugins to skip assets compilation with Go template


vianhazman:
Some plugins might have their own assets compilation method for example using Go templates or Jinja which used their own set of variables. For example an email sending plugin might have various kinds of variables:
```
Dear {{ .RECIPIENTS }},
Attached in this email is the monthly report for {{ .MONTH }}
```
Based on the discussion with @kushsharma , the proposed workaround is to add `SkipCompile` flag into `CompileAssetsResponse` which is configurable at plugin level.

PR #75, opened 2021-08-30
feat(plugin): Allow plugins to skip assets compilation with Go template

vianhazman:
Closes https://github.com/odpf/optimus/issues/74


instance/context.go
```diff
@@ -301,18 +305,22 @@ func DumpAssets(jobSpec models.JobSpec, scheduledAt time.Time, engine models.Tem
...
 	// compile again if needed
-	templates, err := engine.CompileFiles(assetsToDump, map[string]interface{}{
-		ConfigKeyDstart:        jobSpec.Task.Window.GetStart(scheduledAt).Format(models.InstanceScheduledAtTimeLayout),
-		ConfigKeyDend:          jobSpec.Task.Window.GetEnd(scheduledAt).Format(models.InstanceScheduledAtTimeLayout),
-		ConfigKeyExecutionTime: scheduledAt.Format(models.InstanceScheduledAtTimeLayout),
-		ConfigKeyDestination:   jobDestination,
-	})
-	if err != nil {
-		return nil, errors.Wrap(err, "failed to compile templates")
+	if !skipCompile {
```
kushsharma, line 312: `DumpAssets` is used during dryrun of the job in validation step. I am afraid if during that stage it is not compiled properly, validation will fail. We will need to skip validation of these specs somehow

ravisuhag:
Asset compilation is not needed at the moment. This can be considered in the future though. Closing the issue at the moment.

PR #80, opened 2021-09-08
Cleanup

sravankorumilli:


Makefile
```diff
@@ -7,9 +7,7 @@ LAST_TAG := "$(shell git rev-list --tags --max-count=1)"
 OPMS_VERSION := "$(shell git describe --tags ${LAST_TAG})-next"
 PROTON_COMMIT := "6fa8d49d26bfd7aabc0978de6e9039bb35732695"
 
-all: build
```
kushsharma, line 10: this was added so that the default command starts as `build` when you just run `make`. We can either use 
```
.DEFAULT_GOAL := build
```
but simply removing it doesn't look good to me.

Makefile
```diff
@@ -7,9 +7,7 @@ LAST_TAG := "$(shell git rev-list --tags --max-count=1)"
...
-
-.PHONY: build smoke-test unit-test test clean generate dist init vet
+.PHONY: build smoke-test unit-test test clean generate vet
```
kushsharma, line 10: add `install` here. I think I renamed `init` to `install`

Makefile
```diff
@@ -46,9 +44,6 @@ vet: ## run go vet
 coverage: ## print code coverage
 	go test -race -coverprofile coverage.txt -covermode=atomic ./... -tags=unit_test && go tool cover -html=coverage.txt
 
-dist: generate
-	@bash ./scripts/build-distributables.sh
```
kushsharma, line 50: we can add goreleaser build command here I think

ravisuhag:
@sravankorumilli Can we please resolve these issues to close this PR. Or close it if it is not needed.

sravankorumilli:
resolved, merging the PR

PR #103, opened 2021-10-25
refactor:  project and namespace configuration to support multiple namespaces

arinda-arif:
- Project & Namespace configs in Optimus.yaml.
  - Project and namespace flags should also be there with priority: flag > yaml > env 
- For cases where project configs are empty (Optimus.yaml is in namespace level), project config should not be overridden.
- Replace current configuration library to use Salt

Config example:
```yaml
version: 1
host: sample_host
project:
  name: sample_project
  config:
    environment: sample_environment
    storage_path: sample_path
namespace:
  name: sample_namespace
  config: {}
  job:
    path: jobs
  datastore:
  - type: bigquery
    path: bigquery
    backup:
      ttl: 720h
      prefix: sample_prefix
      dataset: sample_dataset
```


config/config.go
```diff
@@ -1,266 +1,162 @@
...
+	Server    ServerConfig    `mapstructure:"serve"`
+	Log       LogConfig       `mapstructure:"log"`
+	Scheduler SchedulerConfig `mapstructure:"scheduler"`
+	Admin     AdminConfig     `mapstructure:"admin"`
 }
 
 type Datastore struct {
 	// type could be bigquery/postgres/gcs
-	Type string `yaml:"type" koanf:"type"`
+	Type string `yaml:"type" mapstructure:"type" koanf:"type"`
```
arinda-arif, line 32: @zakiindra i think we missed this, koanf still defined

config/config.go
```diff
@@ -1,266 +1,162 @@
...
 
-	DB                      DBConfig       `yaml:"db"`
-	Metadata                MetadataConfig `yaml:"metadata"`
-	ReplayNumWorkers        int            `yaml:"replay_num_workers"`
-	ReplayWorkerTimeoutSecs time.Duration  `yaml:"replay_worker_timeout_secs"`
-	ReplayRunTimeoutSecs    time.Duration  `yaml:"replay_run_timeout_secs"`
+	DB                      DBConfig       `yaml:"db" mapstructure:"db"`
+	Metadata                MetadataConfig `yaml:"metadata" mapstructure:"metadata"`
+	ReplayNumWorkers        int            `yaml:"replay_num_workers" mapstructure:"replay_num_workers" default:"1"`
+	ReplayWorkerTimeoutSecs time.Duration  `yaml:"replay_worker_timeout_secs" mapstructure:"replay_worker_timeout_secs" default:"120s"`
```
arinda-arif, line 79: lets remove `Secs` since now it is duration, so we can have `ReplayWorkerTimeout` and `ReplayRunTimeout`

config/config.go
```diff
@@ -1,266 +1,162 @@
...
+	ReplayRunTimeoutSecs    time.Duration  `yaml:"replay_run_timeout_secs" mapstructure:"replay_run_timeout_secs"`
 }
 
 type DBConfig struct {
 	// database connection string
 	// e.g.: postgres://user:password@host:123/database?sslmode=disable
-	DSN string `yaml:"host"`
+	DSN string `yaml:"host" mapstructure:"dsn"`
```
arinda-arif, line 86: looks like using Salt, `yaml` is not necessary to be defined. lets try to remove it

arinda-arif: tried this and works fine

cmd/deploy.go
```diff
@@ -83,27 +90,22 @@ func postDeploymentRequest(l log.Logger, projectName string, namespace string, j
...
-	if err != nil {
-		return errors.Wrap(err, "failed to update project configurations")
-	} else if !registerResponse.Success {
-		return fmt.Errorf("failed to update project configurations, %s", registerResponse.Message)
+	if len(projectSpec.Config) > 0 {
+		// update project and namespace config if needed
+		registerProject(l, runtime, deployTimeoutCtx, projectSpec, namespaceSpec)
+	} else {
+		// update namespace config if needed
+		registerProjectNamespace(l, runtime, deployTimeoutCtx, projectSpec.Name, namespaceSpec)
```
arinda-arif, line 107: not merging, keeping exactly as in yaml.

logic should be in the server side.

register project and namespace should be done separately

arinda-arif:
Feedback as per discussed with @sravankorumilli @kushsharma 
1. Config Grouping the jobs and datastore under namespace
2. Project and Namespace should be registered separately
3. Validation of project & namespace in server side, not delete namespace configs if empty (logic in server side)
4. Remove validation logic in cli

arinda-arif:
have tested the functionality with @zakiindra. additional note:
- `config init` command also exposes server, scheduler, and all other configs to .optimus.yaml. should check if this is expected? since the usage of this command might be for initializing project repository configs.

@kushsharma need your feedback on these changes 🙂

config/config.go
```diff
@@ -1,281 +1,171 @@
...
-	DSN string `yaml:"host"`
+	DSN string `mapstructure:"dsn"`
 
 	// maximum allowed idle DB connections
-	MaxIdleConnection int `yaml:"max_idle_connection"`
+	MaxIdleConnection int `mapstructure:"max_idle_connection" default:"5"`
 
 	// maximum allowed open DB connections
-	MaxOpenConnection int `yaml:"max_open_connection"`
+	MaxOpenConnection int `mapstructure:"max_open_connection" default:"10"`
```
kushsharma, line 92: I think these numbers were changed to 20(open) and 5(idle)?

config/loader.go
```diff
@@ -27,76 +20,28 @@ var (
...
-	}
-	if os.IsNotExist(err) {
-		return false, nil
-	}
-	return false, err
+	var o Optimus
+
+	currPath, _ := os.Getwd()
+	execPath, _ := os.Executable()
+	currentHomeDir, _ := os.UserHomeDir()
```
kushsharma, line 27: Any reason we are ignoring errors?

store/postgres/namespace_repository.go
```diff
@@ -113,6 +113,9 @@ func (repo *namespaceRepository) Save(ctx context.Context, spec models.Namespace
 	} else if err != nil {
 		return errors.Wrap(err, "unable to find namespace by name")
 	}
+	if len(spec.Config) == 0 {
+		return store.ErrEmptyConfig
+	}
```
kushsharma, line 118: Does this mean we can't save a namespace with empty configs? I don't think this will be nice right?

config/config.go
arinda-arif: updated

config/config.go
arinda-arif: updated

config/config.go
arinda-arif: hi @zakiindra, do you have context on this?

api/handler/v1/runtime.go
```diff
@@ -256,22 +256,10 @@ func (sv *RuntimeServiceServer) RegisterProject(ctx context.Context, req *pb.Reg
 	projectSpec := sv.adapter.FromProjectProto(req.GetProject())
 
 	if err := projectRepo.Save(ctx, projectSpec); err != nil {
-		return nil, status.Errorf(codes.Internal, "%s: failed to save project %s", err.Error(), req.GetProject().GetName())
-	}
-
-	if req.GetNamespace() != nil {
```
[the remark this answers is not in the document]
arinda-arif: discussed with sravan and zaki: will mark the proto as deprecated and add the warning log in Optimus. while all migrated, we can remove from proto.

config/config.go
arinda-arif: also need to cross verify other default values

store/postgres/namespace_repository.go
arinda-arif: Whenever the namespace is not present and config is empty, it will still be created. However, if namespace is already there, it will not update on empty configs.

PR #104, opened 2021-10-28
feat: add extension supports

irainia:
Add feature for Optimus to support extension. This extension works similarly to the one in Github CLI, though there are some differences. The following is the current functionality supported:

* Install

Allows the user to install an extension. Currently, it is limited to the Github repository only. Command to install:

```zsh
optimus extension install [owner/repo]
```

an example:

```zsh
optimus extension install gojek/optimus-extension-valor
```

The above example will install an extension for Optimus named Valor. The convention for a repository to be considered as Optimus' extension is its repo name should have prefix `optimus-extension-`. Currently, installation is only using the repository's asset binary.

After installation, another command will be available in Optimus. An example before installation:

```zsh
optimus
```

results:

```zsh
...
config ...
create ...
deploy ...
...
```

but after installation, based on the example above:

```zsh
...
config ...
create ...
deploy ...
valor ...
...
```

The name of the extension will be available. Though, the user could put alias for it if necessary. **Still, the user cannot use extension with the same name or alias with the default command.**

* Run

Executes an already installed extension. Taking example of Valor, then command to use:

```zsh
optimus valor [args]
```


extension/extension.go
```diff
@@ -0,0 +1,191 @@
...
+
+	reservedCommands []string
+}
+
+// NewExtension initializes extension
+func NewExtension(
+	ctx context.Context, manifest *Manifest,
+	githubClient *github.Client, httpClient *http.Client,
+	reservedCommands ...string,
+) (*Extension, error) {
```
kushsharma, line 41: I don't see this throwing error in any case?

extension/extension.go
```diff
@@ -0,0 +1,191 @@
...
+	return &Extension{
+		ctx:              ctx,
```
kushsharma, line 43: Why do we have `context` as part of the member variable?

cmd/extension.go
```diff
@@ -0,0 +1,103 @@
...
+						return errors.New("one argument for [owner/repo] is required")
+					}
+					splitArg := strings.Split(args[0], "/")
+					if len(splitArg) != 2 || splitArg[0] == "" || splitArg[1] == "" {
+						return errors.New("argument should follow pattern [owner/repo]")
+					}
+
+					owner := splitArg[0]
+					repo := splitArg[1]
+					alias, _ := cmd.Flags().GetString("alias")
```
kushsharma, line 87: what if we use `StringVarP` for this flag, we don't need to worry about fetching it later and making sure there was no error?

extension/extension.go
```diff
@@ -0,0 +1,191 @@
...
+	var path string
+	for _, metadata := range e.manifest.Metadatas {
+		for _, a := range metadata.Aliases {
+			if name == a {
+				path = metadata.LocalPath
+			}
+		}
+	}
+	cmd := exec.Command(path, args...)
+	cmd.Stdout = os.Stdout
```
kushsharma, line 189: I think we should also pipe `Stdin` and `Stderr`

extension/extension.go
```diff
@@ -0,0 +1,191 @@
...
+	return &Extension{
+		ctx:              ctx,
+		githubClient:     githubClient,
+		httpClient:       httpClient,
+		manifest:         manifest,
+		reservedCommands: reservedCommands,
+	}, nil
+}
+
+// Install installs extension from a Gilthub repository
```
kushsharma, line 51: `Gilthub`

extension/manifest.go
```diff
@@ -0,0 +1,77 @@
...
+		return fmt.Errorf("error opening file: %v", err)
+	}
+	defer f.Close()
+
+	_, err = f.Write(content)
+	if err != nil {
+		return fmt.Errorf("error writing file: %v", err)
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
irainia: Metadata contains information about an extension, such as the path where it is located in the user's local machine or alias when calling it from Optimus. With the metadata being known, it will be easy to install and to execute a particular extension. The user can install more than one extension in their local. And each extension requires metadata to be specified. And all of those metadata are stored in a file named Manifest.

extension/extension.go
irainia: The context is required when trying to get the latest release information. I agree that we can create a new context inside the calling method. But, the context is also used when trying to download the latest release binary file. So, rather than re-initializing it, perhaps we can re-use it by sending it through the argument.

cmd/extension.go
irainia: we can do that, but then in the `Install` method, it will receive the arguments like:
```go
... Install(owner, repo string, alias *string) ...
```
instead of:
```go
... Install(owner, repo, alias string) ...
```

Both approaches require checking and I don't think error checking is required here.

extension/io.go
```diff
@@ -0,0 +1,17 @@
+package extension
+
+import (
+	"fmt"
+	"os"
+	"path"
+)
+
+const extensionDirName = ".optimus"
```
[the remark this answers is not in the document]
irainia: Got it. I will change it to `.optimus/extensions` later. This directory will be stored in the user's home directory. Or in other words, it will be in `~/.optimus/extensions`.

extension/extension.go
irainia: You're right, will remove it later.

extension/extension.go
irainia: Okay, will add it later along with the others.

extension/extension.go
irainia: Will update it along with the others.

extension/extension.go
kushsharma: I think it's better if we move it to `Install` as the first argument so the caller can decide to inject the context.

irainia: I have no strong opinion whether it is better in the initializer or in the method since I saw some repos use it in either of them. Sure, I can move it `Install`.

cmd/extension.go
irainia: Oh right, I just realised that it accepts the pointer. I thought that it returns the pointer when calling it. Sure, we can use `StringVarP`.

extension/manifest.go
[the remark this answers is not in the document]
irainia: Yes, at least `alias` is required. And the way we store it, if not in the file, where would you suggest it would be?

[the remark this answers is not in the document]
irainia: Also, later we planned to have an `Upgrade` function, which requires the repository information such as its ID and the current tag. Apart from that, though it's still in consideration, there might be a `List` function where it will list all the installed extensions and also check if there's a possible update those extensions.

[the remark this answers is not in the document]
kushsharma: Alright, that make sense.

cmd/extension.go
```diff
@@ -0,0 +1,100 @@
...
+
+func getUsedCommands(cmd *cli.Command) []string {
+	output := make([]string, len(cmd.Commands()))
+	for i, c := range cmd.Commands() {
+		output[i] = c.Name()
+	}
+	return output
+}
+
+func extensionCommand(installFn func(owner, repo, alias string) error, l log.Logger) *cli.Command {
```
kushsharma, line 62: Can we accept an `interface` here? It will also help write unit tests for the extension package which I don't see any?

irainia: I don't think `interface` is necessary. We can puth any `installFn` function as we want, since it accepts the pointer. As for the `logger`, I believe it's already an interface.

irainia: Oh just realised it, for the extension, yes I didn't write it due to schedule. I will add them later and mention it here.

extension/io.go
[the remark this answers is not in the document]
irainia: It is updated as suggested

extension/extension.go
irainia: After some implementation changes, I decided to keep it but I added some checks which could return errors.

cmd/extension.go
irainia: I restructured the extension command for better readability, and add interface as well as suggested,

PR #117, opened 2021-12-07
fix: priority resolution should not fail because of missing jobs

kushsharma:


arinda-arif:
LGTM. Functionality also runs as expected.

Issue #89, opened 2021-09-26
Two job with same name in different project generate incorrect dependencies


kushsharma:
If the same job name is used in two projects, when the dependencies are resolved in the current project, it should always select the job from the current project and not treat it as cross-project dependency.

kushsharma:
This looks similar to https://github.com/odpf/optimus/issues/41

PR #118, opened 2021-12-07
fix: dependency resolution during duplicate destinations of job

kushsharma:
https://github.com/odpf/optimus/issues/89

if during dependency resolution, multiple jobs have same destination it should give higher priority to select a job that belongs to the same project first, then choose anyone from the rest if not found.

### DIFF
--- a/job/dependency_resolver.go
+++ b/job/dependency_resolver.go
@@ -70,8 +70,8 @@ func (r *dependencyResolver) resolveInferredDependencies(ctx context.Context, jo
 
 	// get job spec of these destinations and append to current jobSpec
 	for _, depDestination := range jobDependencies {
-		depSpec, depProj, err := projectJobSpecRepo.GetByDestination(ctx, depDestination)
-		if err != nil {
+		projectJobPairs, err := projectJobSpecRepo.GetByDestination(ctx, depDestination)
+		if err != nil || len(projectJobPairs) == 0 {
 			if err == store.ErrResourceNotFound {
 				// should not fail for unknown dependency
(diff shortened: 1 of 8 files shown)


job/dependency_resolver.go
```diff
@@ -70,30 +70,62 @@ func (r *dependencyResolver) resolveInferredDependencies(ctx context.Context, jo
...
-	if dependency.Project.Name == currentJobSpecProject {
-		return models.JobSpecDependencyTypeIntra
+// extractDependency extracts tries to find the upstream dependency from multiple matches
+// type of dependency is decided based on if the job belongs to same project or other
+// Note(kushsharma): correct way to do this is by creating a unique destination for each job
+// this will require us to either change the database schema or add some kind of
+// unique literal convention
+func extractDependency(projectJobPairs []store.ProjectJobPair, projectSpec models.ProjectSpec) models.JobSpecDependency {
+	var dep models.JobSpecDependency
+	if len(projectJobPairs) == 1 {
```
arinda-arif, line 96: is this needed? or we can just skip this, and go straight to check if any job within the same project exists or not. that way we can remove the duplication of getting the first element and remove the logic of checking inter dependency I believe.

kushsharma: Yes, I agree this will simplify the code but will ignore the fact that this should not happen, we should not have this ambiguity in the first place. If you read the comment on line 110, I wanted to point out that this is not a perfect way to resolve this. Ideally, we should always only get one job for one destination.

arinda-arif: makes sense 👍

arinda-arif:
LGTM. Functionality also runs as expected.

PR #120, opened 2021-12-09
feat: resource config

irainia:
Changes:
* allows the user to configure the resource for job
* additional `metadata` field under `job.yaml` file
* when creating job, the user has the option to configure the resource

Example:
```yaml
version: 1
name: allowed_cat
owner: anwar@odpf
schedule:
  start_date: "2021-12-08"
  interval: 0 2 * * *
...
metadata:
  resource:
    request:
      memory: 128Mi
      cpu: 250m
    limit:
      memory: 256Mi
      cpu: 500m

```


ext/scheduler/airflow2/compiler/compiler.go
```diff
@@ -48,6 +48,11 @@ func (com *Compiler) Compile(schedulerTemplate []byte, namespaceSpec models.Name
 		}
 	}
 
+	var metadata *models.JobSpecMetadata
+	if jobSpec.Metadata != (models.JobSpecMetadata{}) {
+		metadata = &jobSpec.Metadata
+	}
```
kushsharma, line 54: what is this check exactly? Are we checking for `nil`?

kushsharma (commented):
I see you have not updated test cases to accommodate new changes, any specific reason?

kushsharma:
You should also update this in docs

ext/scheduler/airflow2/compiler/compiler.go
irainia: The reason I use the address instead of the values is to ensure that if it's empty, it won't get compiled at all in the template. The check in line 52 is basically checking if the `jobSpec.Metadata` is not default value (meaning it's not empty), then I will assign the address to the `metadata` shown by line 53.

irainia:
> You should also update this in docs

Sure, will add the docs

ext/scheduler/airflow2/compiler/compiler.go
kushsharma: Oh, but what if metadata is not empty but it doesn't have resources configs?

irainia: Well, good point. It should've been for the resource only. I'll update it.

irainia:
> > You should also update this in docs
> 
> Sure, will add the docs

docs are committed, which contains the  resource and another topic related to extension

ext/scheduler/airflow2/compiler/compiler.go
irainia: It's updated

cmd/create.go
```diff
@@ -224,6 +224,14 @@ Note: remove interval field from job specification for manually triggered jobs`,
...
 			},
 		},
+		{
+			Name: "metadata.resource",
+			Prompt: &survey.Confirm{
+				Message: "Do you want to configure resource?",
+				Default: false,
+				Help:    "Set configuration for CPU or Memory for each job",
+			},
+		},
```
kushsharma, line 234: I don't think we should ask this as a question. This will be mostly used by advanced users and adding this in normal flow will only add confusion.

ext/scheduler/airflow2/resources/base_dag.py
```diff
@@ -53,6 +53,38 @@
     {{ base $baseTaskSchema.SecretPath | quote }}
 )
 {{- end }}
+
```
kushsharma, line 56: All of this should be added in `ext/scheduler/airflow/...` as well to support airflow 1.

docs/docs/guides/create-job.md
```diff
@@ -107,8 +126,18 @@ labels:
    orchestrator: optimus
 dependencies: []
 hooks: []
+metadata:
```
kushsharma, line 129: Can you also add this with 1-2 description on [this](https://odpf.github.io/optimus/docs/concepts/overview#job-specification) section?

cmd/create.go
irainia: So, are you suggesting to not show this question? Like, the user can configure it by manually updating the `job.yaml` file?

ext/scheduler/airflow2/resources/base_dag.py
irainia: Oh right, I haven't put it. Will update it.

docs/docs/guides/create-job.md
irainia: sure, updating

cmd/create.go
kushsharma: Yes, they can manually add that in yaml.

irainia: it's updated

ext/scheduler/airflow2/resources/base_dag.py
irainia: it's updated

docs/docs/guides/create-job.md
irainia: it's updated

PR #122, opened 2021-12-10
feat: allow manually triggered job using empty scheduler interval

kushsharma:


arinda-arif:
LGTM. Job spec with empty interval successfully compiled to dag with schedule_interval `NONE`. Jobs with schedule are also compiled as usual.

PR #127, opened 2021-12-13
docs: modify optimus configuration sample

arinda-arif:
As part of supporting multiple namespaces in a project, there were breaking changes in the Optimus project configuration.

### DIFF
--- a/docs/docs/getting-started/configuration.md
+++ b/docs/docs/getting-started/configuration.md
@@ -10,22 +10,35 @@ version: 1
 # used to connect optimus service
 host: localhost:9100 
 
-jobs:
-  # folder where job specifications are stored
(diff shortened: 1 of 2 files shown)


docs/docs/getting-started/configuration.md
```diff
@@ -10,22 +10,35 @@ version: 1
...
+    # folder where job specifications are stored
+    path: "job"
+    
+  datastore:
+    # optimus is capable of supporting multiple datastores
+    type: bigquery
+    # path where resource spec for BQ are stored
+    path: "bq"
+    # backup configurations of a datastore
+    backup: {}
```
kushsharma, line 38: we should add available optional config options here

PR #129, opened 2021-12-17
feat: create and update secrets through cli & api

arinda-arif:


cmd/secret.go
```diff
@@ -0,0 +1,240 @@
...
+	return secretCmd
+}
+
+func getSecretName(args []string) (string, error) {
+	if len(args) < 1 {
+		return "", errors.New("secret name is required")
+	}
+	if strings.HasPrefix(args[0], models.SecretTypeSystemDefinedPrefix) {
+		return "", errors.New(fmt.Sprintf("secret name cannot be started with %s", models.SecretTypeSystemDefinedPrefix))
+	}
```
kushsharma, line 132: we should have this check server side

models/project.go
```diff
@@ -21,6 +21,12 @@ const (
 
 	// Secret used to authenticate with scheduler provided at ProjectSchedulerHost
 	ProjectSchedulerAuth = "SCHEDULER_AUTH"
+
+	SecretTypeSystemDefined SecretType = "system_defined"
```
kushsharma, line 25: no need for `defined` in the literal I think

store/postgres/migrations/000016_update_replay_table_add_config_column.down.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE replay DROP IF EXISTS config;
```
kushsharma, line 1: Why is replay table being modified in this PR?

kushsharma:
I was thinking what if we have a single `set` command instead of `create/update`?
@sravankorumilli @arinda-arif

cmd/secret.go
arinda-arif: @kushsharma we are not doing it on the server side because people should be able to create user-defined and system-defined secrets through API, while through CLI it is only user-defined secrets.

Sure we can still validate it on the server-side, but we may need to add a field to differentiate those 2 secret types. what do you think?

store/postgres/migrations/000016_update_replay_table_add_config_column.down.sql
arinda-arif: it should not. this is a new migration file that is being added in other PR (not merged yet), and we are adding 000017 in this PR. this does not belong to this PR

arinda-arif:
> I was thinking what if we have a single `set` command instead of `create/update`? @sravankorumilli @arinda-arif

I think having create/update gives benefits of having extra validation, for example:
- prevent users from replacing existing secrets by mistake when creating
- prevent creating unnecessary secrets when updating a wrong secret name by mistake

cmd/secret.go
kushsharma: Make sense, let's keep it in CLI only.

kushsharma:
> > I was thinking what if we have a single `set` command instead of `create/update`? @sravankorumilli @arinda-arif
> 
> I think having create/update gives benefits of having extra validation, for example:
> 
> * prevent users from replacing existing secrets by mistake when creating
> * prevent creating unnecessary secrets when updating a wrong secret name by mistake
> 
> And the flexibility to differentiate who is authorized to create or update in the future.

Make sense, what if we solve this using a flag `--update-only`? I am trying to make APIs a little more simpler. What do you think?

arinda-arif:
> > > I was thinking what if we have a single `set` command instead of `create/update`? @sravankorumilli @arinda-arif
> > 
> > 
> > I think having create/update gives benefits of having extra validation, for example:
> > 
> > * prevent users from replacing existing secrets by mistake when creating
> > * prevent creating unnecessary secrets when updating a wrong secret name by mistake
> > 
> > And the flexibility to differentiate who is authorized to create or update in the future.
> 
> Make sense, what if we solve this using a flag `--update-only`? I am trying to make APIs a little more simpler. What do you think?

Got your point. I assume this is the behavior if we are having that flag:
* `optimus create secret` -> will create a secret. if any existing secret is found, will not be updated.
* `optimus create secret --update-only` -> will update a secret. if the secret is not found, will not be created.

Does `create secret --update-only` seems odd? should we use `register` instead or is there any other suggestion? 
cc @sbchaos @sravankorumilli

sbchaos:
> Does `create secret --update-only` seems odd? should we use `register` instead or is there any other suggestion? cc @sbchaos @sravankorumilli

To me it sounds a bit odd that we are using create to update the value of secret. Register seems like a reasonable alternative to update/create.

ravisuhag:
+1 to `set` as suggested by @kushsharma

arinda-arif:
Agree, let's have `set` then. For the flag, should we go with `--update`? having `--update-only` seems to skip the flag means it can be either create/update, while we actually want to separate both behaviors to avoid the above issues.

cmd/secret.go
```diff
@@ -0,0 +1,160 @@
...
+)
+
+var (
+	secretTimeout = time.Minute * 15
+)
+
+func secretCommand(l log.Logger, conf config.Provider) *cli.Command {
+	cmd := &cli.Command{
+		Use:   "secret",
+		Short: "manage secrets to be used in jobs",
```
kushsharma, line 28: Capitalize the first character "Manage"

cmd/secret.go
```diff
@@ -0,0 +1,160 @@
...
+	encoded := false
+	updateOnly := false
+
+	var (
+		projectName   string
+		namespaceName string
+		filePath      string
+	)
+
+	secretCmd := &cli.Command{
```
kushsharma, line 44: We can merge these lines

cmd/secret.go
```diff
@@ -0,0 +1,160 @@
...
+		Use:     "set",
+		Short:   "set secret",
+		Example: "optimus secret set sample_secret secret_value",
```
kushsharma, line 47: `optimus secret set <secret_name> <secret_value>`

cmd/secret.go
```diff
@@ -0,0 +1,160 @@
...
+
+	var (
+		projectName   string
+		namespaceName string
+		filePath      string
+	)
+
+	secretCmd := &cli.Command{
+		Use:     "set",
+		Short:   "set secret",
```
kushsharma, line 46: We should add a small one-line description here

cmd/secret.go
```diff
@@ -0,0 +1,160 @@
...
+This operation takes secret name as its first argument. 
+Secret value can be either provided in second argument or through file flag. 
+Use base64 flag if the value has been encoded.
+		`,
+	}
+	secretCmd.Flags().StringVarP(&projectName, "project", "p", conf.GetProject().Name, "project name of optimus managed repository")
+	secretCmd.Flags().StringVarP(&namespaceName, "namespace", "n", conf.GetNamespace().Name, "namespace of deployee")
+	secretCmd.Flags().BoolVarP(&encoded, "base64", "", encoded, "create secret with value that has been encoded")
+	secretCmd.Flags().BoolVarP(&updateOnly, "update-only", "", updateOnly, "only update existing secret, do not create new")
+	secretCmd.Flags().StringVarP(&filePath, "file", "f", filePath, "provide file path to create secret from file instead")
```
kushsharma, line 58: Please make the first character of all the description capital.

cmd/secret.go
```diff
@@ -0,0 +1,160 @@
...
+		Long: `
+This operation takes secret name as its first argument. 
+Secret value can be either provided in second argument or through file flag. 
+Use base64 flag if the value has been encoded.
+		`,
+	}
+	secretCmd.Flags().StringVarP(&projectName, "project", "p", conf.GetProject().Name, "project name of optimus managed repository")
+	secretCmd.Flags().StringVarP(&namespaceName, "namespace", "n", conf.GetNamespace().Name, "namespace of deployee")
+	secretCmd.Flags().BoolVarP(&encoded, "base64", "", encoded, "create secret with value that has been encoded")
+	secretCmd.Flags().BoolVarP(&updateOnly, "update-only", "", updateOnly, "only update existing secret, do not create new")
```
kushsharma, line 57: use BoolVar

docs/docs/guides/manage-secrets.md
```diff
@@ -0,0 +1,51 @@
...
+```shell
+$ optimus secret set someSecret --file=/path/to/secret
+```
+
+Please note that registering a secret which already exists will result in error. Modifying an existing secret can be 
+done using the Update command.
+
+## Updating a secret
+
+Updating an already existing secret in Optimus can be done passing the flag update-only to set command:
```
kushsharma, line 34: These examples are misleading, `update-only` flag is generally used when you explicitly only want to update a secret when it is already there and don't want to create it by mistake. I think we should write it as such.

store/postgres/migrations/000016_update_replay_table_add_config_column.down.sql
kushsharma: This is still not removed?

store/store.go
```diff
@@ -44,7 +44,8 @@ type ProjectRepository interface {
 
 // ProjectSecretRepository stores secrets attached to projects
 type ProjectSecretRepository interface {
-	Save(ctx context.Context, item models.ProjectSecretItem) error
+	Register(ctx context.Context, item models.ProjectSecretItem) error
```
kushsharma, line 47: We renamed Save to register, any reasoning?

api/handler/v1/runtime.go
```diff
@@ -656,12 +656,30 @@ func (sv *RuntimeServiceServer) RegisterSecret(ctx context.Context, req *pb.Regi
...
+	secretRepo := sv.secretRepoFactory.New(projSpec, namespaceSpec)
+	if req.UpdateOnly {
+		if err := secretRepo.Update(ctx, models.ProjectSecretItem{
+			Name:  req.GetSecretName(),
+			Value: string(base64Decoded),
+		}); err != nil {
+			return nil, status.Errorf(codes.Internal, "%s: failed to update secret %s", err.Error(), req.GetSecretName())
+		}
+	} else {
+		if err := secretRepo.Save(ctx, models.ProjectSecretItem{
```
kushsharma, line 677: wouldn't the current implementation of `Save` will fail this if the secret is already registered? I was thinking the behavior should be 
- if the `set` is called and it doesn't exist, it will create it
- if the `set` is called and it does exist, it will update it
- if the `set` is called with `--update-only` flag and it doesn't exist, it will fail
- if the `set` is called with `--update-only` flag and it does exist, it will update

cmd/secret.go
```diff
@@ -0,0 +1,159 @@
...
+	pb "github.com/odpf/optimus/api/proto/odpf/optimus"
+	"github.com/odpf/optimus/config"
+	"github.com/odpf/salt/log"
+	"github.com/pkg/errors"
+	cli "github.com/spf13/cobra"
+	"google.golang.org/grpc"
+)
+
+var (
+	secretTimeout = time.Minute * 15
```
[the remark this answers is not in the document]
sbchaos: Sure, will update it to 2 minutes

api/handler/v1/runtime.go
sbchaos: The current behaviour will update an existing secret at time of create.
If this is not going to be a problem, then I can keep the current behaviour.

kushsharma: Sorry didn't get what you said, can you rephrase it, please? Is the current behavior match with those 4 points?

sbchaos: Existing behaviour (without change on this PR), save will create a new secret, if it exists, it will update the secret.
It might be a problem when the update of existing secret is unintentional at the time of create.

If we encounter this problem very less(unintentional update) then we can retain the existing behaviour.
Otherwise it will be better to be intentional about create or update of secrets.

api/handler/v1beta1/runtime.go
```diff
@@ -639,35 +638,6 @@ func (sv *RuntimeServiceServer) GetWindow(ctx context.Context, req *pb.GetWindow
 	}, nil
 }
 
-func (sv *RuntimeServiceServer) RegisterSecret(ctx context.Context, req *pb.RegisterSecretRequest) (*pb.RegisterSecretResponse, error) {
```
sbchaos, line 642: Moving secrets management to separate file as runtime file is currently very big.
We can split it later as part of a refactor to jobs, resources and different apis.

cmd/secret.go
```diff
@@ -0,0 +1,234 @@
...
+}
+
+func registerSecret(l log.Logger, conf config.Provider, req *pb.RegisterSecretRequest) (err error) {
+	dialTimeoutCtx, dialCancel := context.WithTimeout(context.Background(), OptimusDialTimeout)
+	defer dialCancel()
+
+	var conn *grpc.ClientConn
+	if conn, err = createConnection(dialTimeoutCtx, conf.GetHost()); err != nil {
+		if errors.Is(err, context.DeadlineExceeded) {
+			l.Info("can't reach optimus service")
```
arinda-arif, line 169: let's use `ErrServerNotReachable`?

cmd/secret.go
```diff
@@ -0,0 +1,234 @@
...
+	secretRequestTimeout, secretRequestCancel := context.WithTimeout(context.Background(), secretTimeout)
+	defer secretRequestCancel()
+
+	l.Info("please wait...")
+	runtime := pb.NewRuntimeServiceClient(conn)
+
+	registerSecretResponse, err := runtime.RegisterSecret(secretRequestTimeout, req)
+	if err != nil {
+		if errors.Is(err, context.DeadlineExceeded) {
+			l.Info("secret registration took too long, timing out")
```
arinda-arif, line 184: We can use color failure for this and capitalize the first letter?

cmd/secret.go
```diff
@@ -0,0 +1,234 @@
...
+		}
+		return errors.Wrapf(err, "request failed for creating secret %s", req.SecretName)
+	}
+
+	if registerSecretResponse.Success {
+		l.Info("secret registered")
```
arinda-arif, line 190: we can also use color success for this and capitalize the first letter

cmd/secret.go
```diff
@@ -0,0 +1,234 @@
...
+		if errors.Is(err, context.DeadlineExceeded) {
+			l.Error(coloredError("Secret update took too long, timing out"))
+		}
+		return errors.Wrapf(err, "Request failed for updating secret %s", req.SecretName)
+	}
+
+	if updateSecretResponse.Success {
+		l.Info(coloredSuccess("Secret updated"))
+	} else {
+		return errors.New(fmt.Sprintf("Request failed for updating secret %s: %s", req.SecretName,
```
kushsharma, line 229: You should not start error with a capital letter as it could be embedded part of some other error. This is done in couple of places in this file

cmd/secret.go
```diff
@@ -0,0 +1,234 @@
...
+
+	updateSecretResponse, err := runtime.UpdateSecret(secretRequestTimeout, req)
+	if err != nil {
+		if errors.Is(err, context.DeadlineExceeded) {
+			l.Error(coloredError("Secret update took too long, timing out"))
+		}
+		return errors.Wrapf(err, "Request failed for updating secret %s", req.SecretName)
+	}
+
+	if updateSecretResponse.Success {
```
kushsharma, line 226: In what case this could be false?

cmd/secret.go
```diff
@@ -0,0 +1,234 @@
...
+			l.Error(ErrServerNotReachable(conf.GetHost()).Error())
+		}
+		return err
+	}
+	defer conn.Close()
+
+	secretRequestTimeout, secretRequestCancel := context.WithTimeout(context.Background(), secretTimeout)
+	defer secretRequestCancel()
+
+	l.Info("please wait...")
```
kushsharma, line 178: Should we use a spinner here? and in Update call?

store/postgres/secret_repository.go
```diff
@@ -84,13 +106,24 @@ func (repo *secretRepository) Insert(ctx context.Context, resource models.Projec
...
 func (repo *secretRepository) Save(ctx context.Context, spec models.ProjectSecretItem) error {
-	existingResource, err := repo.GetByName(ctx, spec.Name)
+	_, err := repo.GetByName(ctx, spec.Name)
 	if errors.Is(err, store.ErrResourceNotFound) {
 		return repo.Insert(ctx, spec)
 	} else if err != nil {
 		return errors.Wrap(err, "unable to find secret by name")
 	}
-	resource, err := Secret{}.FromSpec(spec, repo.project, repo.hash)
+	return errors.New("secret already exist")
```
kushsharma, line 115: We should generalize this error to `resource already exists` or something similar to `store.ErrResourceNotFound` or should get rid of this logic of checking if it exists and only insert the secret? Can you check if gorm throws an error if it already exists?
Because currently, I think we are leaking our implementation detail out of an interface.

cmd/secret.go
sbchaos: Changed the error messages to lowercase

cmd/secret.go
sbchaos: This check is not required.

cmd/secret.go
sbchaos: Using spinner

store/postgres/secret_repository.go
sbchaos: Gorm will throw an error like 
```
\x1b[31;1m /../optimus/store/postgres/secret_repository.go:135 \x1b[35;1mERROR: duplicate key value violates unique constraint \"secret_project_id_name_key\" (SQLSTATE 23505)" ....(prints whole query with values) 
```

Which will return the implementation (and query) details back in the api, so we will need to replace the returned error.

Can you please explain how are we leaking implementation details out of interface ?

sravankorumilli: checking explicitly is fine & we don't need to rely on gorm error.

sbchaos:
Merging the PR, we can still have the conversation if needed.

Any required changes will be made to the followup PR.

PR #130, opened 2022-01-02
feat: progress bar during specification deployment in cli

kushsharma:
- reorganized commands to follow the same convention across CLI
- added spinner and verbose flags in deploy and validate steps for better UX
- `optimus validate job` command is changed to `optimus job validate`


cmd/backup_resource.go
```diff
@@ -0,0 +1,228 @@
...
+	"github.com/odpf/salt/log"
+	"github.com/pkg/errors"
+	cli "github.com/spf13/cobra"
+	"google.golang.org/grpc"
+)
+
+func backupResourceSubCommand(l log.Logger, datastoreRepo models.DatastoreRepo, conf config.Provider) *cli.Command {
+	var (
+		backupCmd = &cli.Command{
+			Use:   "resource",
```
ravisuhag, line 20: @kushsharma IMO, `backup` command should take resource as arg instead of sub command. If user does not pass  resource then we can promote them. 
`backup <id>` WDYT?

ravisuhag: Another way could be to call it `run` instead of `resource`.

kushsharma: term `resource` is currently used for BQ table, Postgres table, GCS bucket, etc whereas `run` will be used for job instances. I think I am confused about what you trying to say, can you elaborate a little more?

ravisuhag: `backup resource` sounds like two nous.

if we are treating backup as noun. then command can be `backup create`  or `backup run`

kushsharma: Makes perfect sense, changed it `optimus backup create --resource "resource_name"`

README.md
```diff
@@ -32,26 +32,28 @@ Optimus has two components, Optimus service that is the core orchestrator instal
...
-optimus is a scaffolding tool for creating transformation job specs
+Optimus is an easy-to-use, reliable, and performant workflow orchestrator for
+data transformation, data modeling, pipelines, and data quality management.
 
 Usage:
   optimus [command]
 
 Available Commands:
+  backup      Backup a resource and its downstream
+  completion  generate the autocompletion script for the specified shell
```
arinda-arif, line 43: Generate supposed to be capital?

cmd/backup.go
```diff
@@ -26,415 +17,13 @@ func backupCommand(l log.Logger, datastoreRepo models.DatastoreRepo, conf config
 	cmd := &cli.Command{
 		Use:   "backup",
 		Short: "Backup a resource and its downstream",
-		Long:  "Backup supported resource of a datastore and all of its downstream dependencies",
-	}
-	cmd.AddCommand(backupResourceSubCommand(l, datastoreRepo, conf))
-	cmd.AddCommand(backupListSubCommand(l, datastoreRepo, conf))
-	cmd.AddCommand(backupDetailSubCommand(l, datastoreRepo, conf))
+		Long: `Backup supported resource of a datastore and all of its downstream dependencies.
+Operation can take upto few minutes to complete. It is advised to check the operation status
```
arinda-arif, line 21: `up to`?

README.md
kushsharma: `completion` command is not added by us and is added by the cobra library so the description is pre-generated. Didn't find an easy way to modify it so I think this looks good enough for now?

cmd/backup.go
kushsharma: Yes, fixed

ravisuhag (approved):
LGTM

cmd/backup_create.go
```diff
@@ -0,0 +1,261 @@
...
+	cli "github.com/spf13/cobra"
+	"google.golang.org/grpc"
+)
+
+func backupCreateCommand(l log.Logger, datastoreRepo models.DatastoreRepo, conf config.Provider) *cli.Command {
+	var (
+		backupCmd = &cli.Command{
+			Use:     "create",
+			Short:   "Create a backup",
+			Example: "optimus backup create --resource <sample_resource_name>",
```
arinda-arif, line 23: in backup users need to put the resource name in a flag, while for replay it is an argument:
- optimus backup create --resource <resource_name>
- optimus replay create <job_name>
should we also make this part consistent?

cmd/deploy.go
```diff
@@ -155,24 +165,30 @@ func postDeploymentRequest(l log.Logger, projectName string, namespaceName strin
...
+					if verbose {
+						// ordinary progress event
+						l.Info(fmt.Sprintf("info '%s': %s", resp.GetResourceName(), resp.GetMessage()))
+					}
 				}
 			}
+			spinner.Stop()
 		}
-		l.Info("deployed resources")
+		l.Info(coloredSuccess("> Resources deployed."))
```
arinda-arif, line 184: Is it intended that we are showing number of jobs deployed but not number of resources deployed?

- Resources deployed.
- Successfully deployed %d/%d jobs

cmd/resource.go
```diff
@@ -0,0 +1,155 @@
...
+				Version:   1,
+				Name:      resourceName,
+				Type:      models.ResourceType(resourceType),
+				Datastore: datastore,
+				Assets:    typeController.DefaultAssets(),
+			}, resourceDirectory); err != nil {
+				return err
+			}
+
+			l.Info(fmt.Sprintf("resource created successfully %s", resourceName))
```
arinda-arif, line 115: add color success and make first letter capital?

README.md
arinda-arif: yes 👍

cmd/resource.go
kushsharma: Fixed it.

cmd/deploy.go
kushsharma: Good catch, fixed it.

PR #133, opened 2022-01-07
fix: rerunning an old job cause task and hook execution time to differ

kushsharma:


sravankorumilli:
As discussed, instead of storing the `execution_time` at task instance level let's just store it at job_run as that is our intention.
That means `instance` table will just be used for logging purpose.

`execution_time` here will means the first time job has been scheduled by the scheduler for a given run, it doesn't change on re runs, when working on the documentation lets capture that too. cc : @arinda-arif

models/job.go
```diff
@@ -106,7 +106,7 @@ type JobSpecNotifier struct {
 }
 
 type JobSpecTask struct {
-	Unit     *Plugin
+	Unit     *Plugin `json:"-" yaml:"-"`
```
sravankorumilli, line 109: what scenarios this change will impact?

kushsharma: Test file where we marshal and unmarshal specs.

sravankorumilli:
changes look good to me

models/job.go
sravankorumilli: Alright

PR #141, opened 2022-01-25
feat: add api to list secrets

sbchaos:


store/postgres/secret_repository.go
```diff
@@ -86,16 +86,46 @@ func (p Secret) ToSpec(hash models.ApplicationKey) (models.ProjectSecretItem, er
...
+		return models.SecretItemInfo{}, err
+	}
+
+	// decrypt secret
+	cleartext, err := cryptopasta.Decrypt(encrypted, hash.GetKey())
+	if err != nil {
+		return models.SecretItemInfo{}, err
+	}
+
+	digest := cryptopasta.Hash("secret", cleartext)
```
kushsharma, line 102: As far as I remember, in RFC we mentioned digest should be generated from encrypted secret not plain text otherwise there are chances of exploitation.

store/postgres/secret_repository.go
```diff
@@ -155,27 +185,29 @@ func (repo *secretRepository) GetByID(ctx context.Context, id uuid.UUID) (models
...
-func (repo *secretRepository) GetAll(ctx context.Context) ([]models.ProjectSecretItem, error) {
-	var specs []models.ProjectSecretItem
+func (repo *secretRepository) GetAll(ctx context.Context) ([]models.SecretItemInfo, error) {
+	var secretItems []models.SecretItemInfo
 	var resources []Secret
-	if err := repo.db.WithContext(ctx).Find(&resources).Error; err != nil {
-		return specs, err
+	if err := repo.db.WithContext(ctx).Preload("Namespace").
+		Joins("LEFT JOIN namespace ON secret.namespace_id = namespace.id").
+		Where("secret.project_id = ? and secret.type = 'user'", repo.project.ID).Find(&resources).Error; err != nil {
```
kushsharma, line 193: we should use enum here for secret type instead of hardcoded string `user`

sbchaos: Sure, will use the enum

store/postgres/secret_repository.go
sbchaos: The question we had was for time, when the app key changes, at that time all the secret digests will change.
The hash function we are using is SHA512 which cannot be easily broken by brute-force attacks.
cc @sravankorumilli

sravankorumilli: Need to check on the rainbow dictionary attack, but I agree it would be good to generate hashes on the encrypted secrets

api/handler/v1beta1/secret.go
```diff
@@ -89,3 +76,21 @@ func getDecodedSecret(encodedString string) (string, error) {
 	}
 	return string(base64Decoded), nil
 }
+
+func mapToGRPCErr(err error, msg string) error {
```
sravankorumilli, line 80: this is nothing related to secret can we move this ato a seperate file

service/namespace_service.go
```diff
@@ -0,0 +1,51 @@
...
+	return &namespaceService{
+		projectService:   projectService,
+		namespaceRepoFac: factory,
+	}
+}
+
+// GetProjectAndNamespace This function is inefficient, it gets the project from repo along with secrets
+// Then we only use the id from the project to fetch the Namespace, along with project and secrets
+// Repository can provide a method to query both together.
+func (s namespaceService) GetProjectAndNamespace(ctx context.Context, projectName string, namespaceName string) (models.ProjectSpec, models.NamespaceSpec, error) {
```
sravankorumilli, line 33: It should just return namespace spec & the callers currently can refer to project spec inside namespace spec right?

sravankorumilli: main concern is namespace service returning projectspec  directly

service/secret_service.go
```diff
@@ -0,0 +1,84 @@
...
+}
+
+type SecretRepoFactory interface {
+	New(projectSpec models.ProjectSpec) store.ProjectSecretRepository
+}
+
+type secretService struct {
+	projService   ProjectService
+	nsService     NamespaceService
+	secretRepoFac SecretRepoFactory
```
sravankorumilli, line 24: It should be initialized with the secretRepo instead of factory directly right?

store/postgres/secret_repository.go
```diff
@@ -86,56 +85,81 @@ func (p Secret) ToSpec(hash models.ApplicationKey) (models.ProjectSecretItem, er
...
+func (repo *secretRepository) Save(ctx context.Context, namespace models.NamespaceSpec, spec models.ProjectSecretItem) error {
 	_, err := repo.GetByName(ctx, spec.Name)
 	if errors.Is(err, store.ErrResourceNotFound) {
-		return repo.Insert(ctx, spec)
+		return repo.Insert(ctx, namespace, spec)
 	} else if err != nil {
 		return errors.Wrap(err, "unable to find secret by name")
 	}
-	return errors.New("secret already exist")
+	return store.ErrResourceExists
```
sravankorumilli, line 139: we can log the secret with the name already exists, as the error that we are returning is generic error

store/postgres/secret_repository.go
```diff
@@ -86,56 +85,81 @@ func (p Secret) ToSpec(hash models.ApplicationKey) (models.ProjectSecretItem, er
...
+	resource, err := Secret{}.FromSpec(spec, repo.project, namespace, repo.hash)
 	if err != nil {
 		return err
 	}
-	if err == nil {
-		resource.ID = existingResource.ID
-	}
+
+	resource.ID = existingResource.ID
+	// TODO: Check project_id at time of update, currently getByName protects from that
```
sravankorumilli, line 154: did you mean validation around project_id being present?

store/postgres/secret_repository.go
```diff
@@ -86,56 +85,81 @@ func (p Secret) ToSpec(hash models.ApplicationKey) (models.ProjectSecretItem, er
 	}, nil
 }
 
+func (p Secret) ToSecretItemInfo(hash models.ApplicationKey) (models.SecretItemInfo, error) {
```
sravankorumilli, line 88: doesn't use hash anymore can be removed, can't we name it as key instead of hash in other places.
We will move the logic of encryption & decryption to service layer can you add that as todo

store/postgres/secret_repository.go
sbchaos: Will remove the Todo.

store/postgres/secret_repository.go
sbchaos: Will remove the param, and rename in other places. Sure will add todo

sravankorumilli (approved):

PR #156, opened 2022-02-21
feat: add labels on job spec as tags in airflow2 dag

novanxyz:
Hi Maintainers,

I need to organizing my dags/jobs on the Airflow UI,
there's already that feature in airflow using tags
https://airflow.apache.org/docs/apache-airflow/stable/howto/add-dag-tags.html
and there's already labels on job spec

so, I'll need that labels rendered as tags in optimus' rendered dag codes.
I hope my code can be tested and reviewed to give this feature on optimus.

thank you


kushsharma:
Thanks for raising the PR @novanxyz , this looks good. 
@arinda-arif Can you help test for any edge cases(different kinds of label strings) and merge if LGTY as well?

irainia:
Thanks for the PR @novanxyz .

Related to the changes you have done, going forward we will move the `labels` into `metadata` in the `job.yaml` configuration file, like [the following](https://github.com/odpf/optimus/blob/main/store/local/job_spec_adapter.go#L45). The use of `labels` itself for the tagging might be misleading. So, instead, we can add the `tags` under `metadata` and use it during rendering.

kushsharma:
@irainia Good point, but can you elaborate on `The use of labels itself for the tagging might be misleading` in airflow dags?

irainia:
Yes @kushsharma , so in the job spec, the user puts values as `label`. However, we actually use the label as `tag` for airflow. I'm thinking that, since we will need an input from the user to be used as a tag in airflow, then why don't we just let them know that they have to specify the `tag` instead of `label`. That way, it will be more consistent from the user's perspective, implementation in the code, and what appears in the airflow.

Regarding whether we will remove the `label` field or not, I have no strong opinion around this. We can still use the `label` field or not. As long as we have some convention like "if it is related monitoring, put it as label, and if it's for tag in airflow, use tag" or something similar, then I'm okay.

kushsharma:
Umm I still feel because we want optimus to be agnostic to airflow as much if we can, adding a `tag` just because airflow uses this terminology won't be better. As per your comment I see the behaviour is still the same, its just a matter of naming and where we put it then I think its better if we just leverage labels for airflow tags. We currently use these labels for GCP resources as well similarly we can do so with airflow. 

Lets hear from others as well @sravankorumilli @arinda-arif @sbchaos

sravankorumilli:
aligned with @kushsharma on not adding anything specific to airflow. Any identifiable information of the job we are expecting users to add it as a label. So, can be used as airflow tags too, it is a similar thought process when its uses in gcp labels.

irainia:
@kushsharma , based on what you mention, you want `labels` to be used for two purposes, which is for the GCP resource labeling and for airflow tagging. Now, how should the user know or the developer know when part of that `label` is considered for airflow or other scheduler (if any) and when used for the GCP resource labelling? An example on the `job.yaml` would be helpful.

And about "agnostic", firstly, I agree that we should be agnostic on the scheduler. But, I'm wondering, how far do we want Optimus to be agnostic to the scheduler provider? Because at the moment, most of our implementation revolves around airflow.

And about naming, I don't think it is "just a matter of naming". I think proper consistency in naming that express the meaning and the purpose of the name should be a good practice, please correct me if I miss it here.

cc: @sravankorumilli

arinda-arif:
i think same labels can be used for anything, either GCP, Airflow, or other scheduler. for example given this job.yaml:
```yaml
labels:
  product: sample_product
```
user can get the total cost for this particular label in GCP billing and can search for dags in Airflow by putting this `product:sample_product` in Airflow UI search bar. 

however, agree that we can move labels to metadata section. how about we use labels for now, and have a separate discussion on handling it in metadata?

ext/scheduler/airflow2/resources/base_dag.py
```diff
@@ -43,7 +43,8 @@
     schedule_interval={{ if eq .Job.Schedule.Interval "" }}None{{- else -}} {{ .Job.Schedule.Interval | quote}}{{end}},
     sla_miss_callback=optimus_sla_miss_notify,
     catchup={{ if .Job.Behavior.CatchUp -}}True{{- else -}}False{{- end }},
-    dagrun_timeout=timedelta(seconds=DAGRUN_TIMEOUT_IN_SECS)
+    dagrun_timeout=timedelta(seconds=DAGRUN_TIMEOUT_IN_SECS),
+    tags=[{{.Job.Labels | values | join '", "' | quote }}]
```
arinda-arif, line 47: job labels is in a `map[string]string` type, not sure it can be joined correctly this way (tested and it failed). how about having similar like this?
```python
tags = [
    {{- range $key, $value := $.Job.Labels}}
    "{{ $key }}:{{ $value }}",
    {{- end}}
]
```
or let me know if you find a better way to do it.

novanxyz: files updated, need team to test it again.
https://github.com/odpf/optimus/pull/156/commits/e7100af1f1212596c6b25adde4c1631c45b493d7

ext/scheduler/airflow/resources/expected_compiled_template.py
```diff
@@ -204,4 +207,4 @@
 
 # arrange failure hook after post hooks
 
-hook_predator >> [ hook_hook__dash__for__dash__fail,]
\ No newline at end of file
+hook_predator >> [ hook_hook__dash__for__dash__fail,]
```
arinda-arif, line 210: @novanxyz also remove the additional new line at the end of this expected file

ext/scheduler/airflow2/resources/expected_compiled_template.py
```diff
@@ -220,4 +223,4 @@
 
 # arrange failure hook after post hooks
 
-hook_predator >> [ hook_hook__dash__for__dash__fail,]
\ No newline at end of file
+hook_predator >> [ hook_hook__dash__for__dash__fail,]
```
arinda-arif, line 226: @novanxyz also remove the additional new line at the end of this expected file. thanks!

Issue #137, opened 2022-01-18
Duplicate cross project dependencies should be handled gracefully


kushsharma:
If a job inferred a cross-project dependency from its task and the same dependency is also mentioned in `job.yaml` specification, they are treated as duplicates. The reason is inferred dependencies when used in map for dependencies uses job name whereas cross-project dependency mentioned in specification uses `project_name/job_name` so duplicates are created inside the same dependency map.
For example:
```yaml
...
dependencies:
- job: foo-project/bar-job
  type: inter
```
The [map](https://github.com/odpf/optimus/blob/0a8d332ea1d989ce7e2fddc0ee4b80be577d3c5b/job/dependency_resolver.go#L165) will have two jobs as `bar-job` and `foo-project/bar-job`.

Although users can choose to simply write the spec properly, the expected result is Optimus to handle it gracefully.

sravankorumilli:
so the optimus compiles it as two sensors right?

kushsharma:
No, actually it generates two of such dependencies
```
d1 >> t1
d1 >> t1
```
but airflow complains of having duplicate order of tasks and simply fails to read the dag.

sravankorumilli:
won't be just addressing this issue alone, lets address entire entire dependency resolution issues as part of https://github.com/odpf/optimus/issues/176

PR #157, opened 2022-02-21
fix: bug caused by inter-dependency miss-assigned map

irainia:
### Description

This PR is related to #137 , and further information is available in the issue.

### Changes

The following is the summary of changes done in this PR:

* update the map key from `depName` into its `jobName`
* prioritised the already found job as the dependency to minimise access to database
* remove unused key in the dependency
* update the unit test

### DIFF
--- a/job/dependency_resolver.go
+++ b/job/dependency_resolver.go
@@ -156,16 +156,18 @@ func (r *dependencyResolver) resolveStaticDependencies(ctx context.Context, jobS
 					}
 					projectName := depParts[0]
 					jobName := depParts[1]
-					if jobSpec.Dependencies[jobName].Job == nil {
-						job, proj, err := projectJobSpecRepo.GetByNameForProject(ctx, projectName, jobName)
-						if err != nil {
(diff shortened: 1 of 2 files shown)


job/dependency_resolver.go
```diff
@@ -156,6 +156,11 @@ func (r *dependencyResolver) resolveStaticDependencies(ctx context.Context, jobS
 					}
 					projectName := depParts[0]
 					jobName := depParts[1]
+
```
sravankorumilli, line 159: this change is not easy to understand, rather lets reconsider entire dependency resolution flow where
1. We move away from user specifiying inter/intra type & the job dependency is inferred automatically.
2. Lets change the scope of this issue or lets consider in the new updated flow of Job Deploy enhancement

irainia: If we want to apply the point (1) you mentioned, then in my opinion, we have need to do the point (2). If not, more conflicts could happend.

sravankorumilli:
Let's address all the concerns around the dependency resolution as a whole as part of  https://github.com/odpf/optimus/issues/176

PR #158, opened 2022-02-21
Namespace name during saving secret is optional

sbchaos:


service/secret_service.go
```diff
@@ -36,13 +36,22 @@ func (s secretService) Save(ctx context.Context, projectName string, namespaceNa
...
 	}
 
-	// TODO: Add new service method to get only project and namespace id for names
-	namespaceSpec, err := s.nsService.Get(ctx, projectName, namespaceName)
-	if err != nil {
-		return err
+	var projectSpec models.ProjectSpec
+	namespaceSpec := models.NamespaceSpec{}
+	var err error
+	if namespaceName == "" { // Namespace is optional for secrets
```
sravankorumilli, line 42: we don't need to fetch the projectSpec & namespaceSpec and do the save, why don't we just pass the namespace name & projectname to the repo layer & repo layer handles saving it by ID.

service/secret_service.go
```diff
@@ -55,12 +64,22 @@ func (s secretService) Update(ctx context.Context, projectName string, namespace
 		return NewError(models.SecretEntity, ErrInvalidArgument, "secret name cannot be empty")
 	}
 
-	namespaceSpec, err := s.nsService.Get(ctx, projectName, namespaceName)
-	if err != nil {
-		return err
+	var projectSpec models.ProjectSpec
+	namespaceSpec := models.NamespaceSpec{}
+	var err error
+	if namespaceName == "" { // Namespace is optional for secrets
```
sravankorumilli, line 70: Same here

service/secret_service.go
sbchaos: It will change the design of entire secret repo, maybe I can do that as part of separate pr.

PR #159, opened 2022-02-22
Enable all linters, disable only the problamatic ones

sbchaos:
We can enable other linters after the issues with them are resolved.


sbchaos:
We want to enable most of the linters, but cannot due to fixes required in code. We will slowly fix those issues and enable the linters. Current list will help us with picking one linter at a time and fixing all the related issues for it.

sravankorumilli (approved):
LGTM

PR #160, opened 2022-02-22
Api to delete secret for project

sbchaos:


service/secret_service.go
```diff
@@ -81,3 +82,17 @@ func (s secretService) List(ctx context.Context, projectName string) ([]models.S
 	}
 	return secretItems, nil
 }
+
+func (s secretService) Delete(ctx context.Context, prjName string, secretName string) error {
```
sravankorumilli, line 86: should consider the namespace name in the request and should use just project name for deleting secrets at project level & namespace name must be used for deleting secrets at namespace level, an extra bit of validation is always safe.

sbchaos: Sure, will add extra validation

sbchaos:
If the secret is not present, it should return error not success.

service/namespace_service.go
```diff
@@ -52,6 +53,28 @@ func (s namespaceService) Get(ctx context.Context, projectName string, namespace
 	return nsSpec, nil
 }
 
+// GetOptional is used for optionally getting namespace if name is present, otherwise get only project
+func (s namespaceService) GetOptional(ctx context.Context, projectName string, namespaceName string) (models.NamespaceSpec, error) {
```
sbchaos, line 57: Return both project and namespace, and name should reflect the namespace can be optional

sravankorumilli (approved):

PR #162, opened 2022-02-24
refactor: Use optimus config struct instead of interface

deryrahman:
- [x] remove config/interface.go
- [x] remove config.Optimus getters


kushsharma:
Hi @deryrahman, although this looks good, can you please help me understand the reasoning behind why we are doing this?

deryrahman:
@kushsharma As of now, since we only have 1 config implementation (which is config.Optimus) we don't need the interface. We also saw that the config is only data structure and have no behavior. And all the fields of conf.Optimus struct is exported (public) so we don't see the urgency of having getter. But if there's a strong reason to use getter, then we can discuss

cmd/server/server.go
```diff
@@ -252,18 +252,18 @@ func jobSpecAssetDump() func(jobSpec models.JobSpec, scheduledAt time.Time) (mod
 	}
 }
 
-func checkRequiredConfigs(conf config.Provider) error {
+func checkRequiredConfigs(conf config.Optimus) error {
```
sbchaos, line 255: Pass only the `config.ServerConfig` to this method.

cmd/secret.go
```diff
@@ -249,14 +249,14 @@ func updateSecret(l log.Logger, conf config.Provider, req *pb.UpdateSecretReques
 	return nil
 }
 
-func listSecret(l log.Logger, conf config.Provider, req *pb.ListSecretsRequest) (err error) {
+func listSecret(l log.Logger, conf config.Optimus, req *pb.ListSecretsRequest) (err error) {
```
sbchaos, line 252: Pass only the `conf.Host` to this method.

cmd/secret.go
```diff
@@ -215,14 +215,14 @@ func registerSecret(l log.Logger, conf config.Provider, req *pb.RegisterSecretRe
 	return nil
 }
 
-func updateSecret(l log.Logger, conf config.Provider, req *pb.UpdateSecretRequest) (err error) {
+func updateSecret(l log.Logger, conf config.Optimus, req *pb.UpdateSecretRequest) (err error) {
```
sbchaos, line 218: Pass only the conf.Host to this method.

cmd/secret.go
```diff
@@ -181,14 +181,14 @@ func validateProperlyEncoded(secretValue string) error {
 	return nil
 }
 
-func registerSecret(l log.Logger, conf config.Provider, req *pb.RegisterSecretRequest) (err error) {
+func registerSecret(l log.Logger, conf config.Optimus, req *pb.RegisterSecretRequest) (err error) {
```
sbchaos, line 184: Pass only the `conf.Host` to this method.

cmd/replay_create.go
```diff
@@ -213,14 +213,14 @@ func printExecutionTree(instance *pb.ReplayExecutionTreeNode, tree treeprint.Tre
 }
 
 func runReplayRequest(l log.Logger, projectName, namespace, jobName, startDate, endDate string, forceRun bool,
-	allowedDownstreamNamespaces []string, conf config.Provider) (string, error) {
```
sbchaos, line 216: Pass only the `conf.Host` to this method.

cmd/replay_create.go
```diff
@@ -108,14 +108,14 @@ Date ranges are inclusive.
 }
 
 func printReplayExecutionTree(l log.Logger, projectName, namespace, jobName, startDate, endDate string,
-	allowedDownstreamNamespaces []string, conf config.Provider) (err error) {
+	allowedDownstreamNamespaces []string, conf config.Optimus) (err error) {
```
sbchaos, line 111: Pass only the `conf.Host` to this method.

cmd/backup_create.go
```diff
@@ -199,14 +199,14 @@ func runBackupDryRunRequest(l log.Logger, conf config.Provider, backupRequest *p
 	return nil
 }
 
-func runBackupRequest(l log.Logger, conf config.Provider, backupRequest *pb.CreateBackupRequest) (err error) {
+func runBackupRequest(l log.Logger, conf config.Optimus, backupRequest *pb.CreateBackupRequest) (err error) {
```
sbchaos, line 202: Pass only the `conf.Host` to this method.

cmd/backup_create.go
```diff
@@ -166,14 +166,14 @@ func extractDescription(description string) (string, error) {
 	return description, nil
 }
 
-func runBackupDryRunRequest(l log.Logger, conf config.Provider, backupRequest *pb.BackupDryRunRequest, backupDownstream bool) (err error) {
+func runBackupDryRunRequest(l log.Logger, conf config.Optimus, backupRequest *pb.BackupDryRunRequest, backupDownstream bool) (err error) {
```
sbchaos, line 169: Pass only the `conf.Host` to this method.

kushsharma:
> Hi @kushsharma, As of now, since we only have 1 config implementation (which is config.Optimus) we don't need the interface. We also saw that the config is only data structure and have no behavior. And all the fields of conf.Optimus struct is exported (public) so we don't see the urgency of having getter. But if there's a strong reason to use getter, then we can discuss. cmiiw @sbchaos @sravankorumilli

Make sense, unless we have more than one implementation it is not needed as of now. The idea of having the interface was if we needed to inject a provider like etcd store or [gcloud runtimevars](https://gocloud.dev/howto/runtimevar/) for runtime changes but I agree that is unlikely in near future.

sbchaos (approved):

PR #194, opened 2022-02-28
refactor: change command for running test in makefile

sbchaos:
* will run unit test only once
* integration test ran all test, will not only run store path
* Integration test setup based on:
https://www.gojek.io/blog/golang-integration-testing-made-easy

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,7 +10,6 @@ PROTON_COMMIT := "2a30976f7b40884ddd90e1792576c0941426e8bc"
 .PHONY: build test test-ci pack-files generate-proto unit-test-ci smoke-test integration-test vet coverage clean install lint
 
 .DEFAULT_GOAL := build
-INTEGRATION_TEST_PATH?=./store
 
 build: pack-files # build optimus binary
 	@echo " > notice: skipped proto generation, use 'generate-proto' make command"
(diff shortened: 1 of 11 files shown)


Makefile
```diff
@@ -32,13 +33,13 @@ generate-proto: ## regenerate protos
 	@echo " > protobuf compilation finished"
 
 unit-test:
-	go list ./... | grep -v -e third_party -e api/proto | xargs go test -count 5 -cover -race -timeout 1m -tags=unit_test
+	go test -count 1 -cover -race -timeout 1m -tags=unit_test ./...
```
kushsharma, line 36: Count `5` was used to try to execute a unit test more than once to catch code with undefined behaviours. I see we have reduced it to `1` here. Any reason we don't want to run these test as before?

Makefile
```diff
@@ -31,18 +30,21 @@ generate-proto: ## regenerate protos
...
+unit-test-ci:
+	go test -count 5 -race -coverprofile coverage.txt -covermode=atomic -timeout 1m -tags=unit_test ./...
 
 smoke-test: build
 	@bash ./scripts/smoke-test.sh
 
-integration-test: 
-	go list ./... | grep -v -e third_party -e api/proto | xargs go test -count 1 -cover -race -timeout 1m
+integration-test:
+	go test -count 1 -cover -race -timeout 1m $(INTEGRATION_TEST_PATH)/...
```
sravankorumilli, line 40: are we going to keep all integration tests in a specific directory or set of directories, I believe they can coexist with other unit tests so we need to run integration tests by configuring right tags.

sbchaos: If we add tag for integration test, we need to annotate all files which are not for integration test with other tag, doc: https://pkg.go.dev/go/build

So using the run directive and rename all test as `TestIntegration`

PR #202, opened 2022-03-04
refactor: refactor and cleanup for secrets management

sbchaos:


store/store.go
```diff
@@ -45,19 +45,20 @@ type ProjectRepository interface {
 
 // ProjectSecretRepository stores secrets attached to projects
 type ProjectSecretRepository interface {
```
sravankorumilli, line 47: secretrepository name is better

service/errors.go
```diff
@@ -71,12 +71,16 @@ func (e *DomainError) Error() string {
 		e.Message, strings.ToLower(e.ErrorType.String()), e.Entity)
 }
 
+func (e *DomainError) Unwrap() error {
+	return e.Err
+}
+
 func (e *DomainError) DebugString() string {
```
sravankorumilli, line 78: need to check on golang has an inbuilt support for this, instead of writing our own.

PR #206, opened 2022-03-08
refactor: fix errors reported by gosimple

sbchaos:


ext/datastore/bigquery/table_spec.go
```diff
@@ -50,11 +50,11 @@ func (t BQTable) FullyQualifiedName() string {
 
 func (t BQTable) Validate() error {
 	switch {
-	case validProjectName.MatchString(t.Project) == false:
```
sbchaos, line 53: Check if we can keep check for false as it is more readable.

sravankorumilli (approved):

PR #213, opened 2022-03-09
refactor: remove airflow1 support

sravankorumilli:


kushsharma:
This will be a major breaking change, any reason we are doing this? Maybe we can stop releasing new features or fixes for airflow1 for now and slowly remove it?

sravankorumilli:
It has been 2 years since airflow 2 is released, until & unless it is being used by anyone there is no need to keep supporting it. If we stop releasing new features or fixes for airflow1 soon latest optimus will become outdated. Let's hear from people, if anyone wants it, yes we can consider providing the support for the same.

cmd/server/server.go
```diff
@@ -292,12 +291,6 @@ func Initialize(l log.Logger, conf config.Optimus) error {
 	jobCompiler := compiler.NewCompiler(conf.Server.IngressHost)
 	// init default scheduler
 	switch conf.Scheduler.Name {
-	case "airflow":
```
sbchaos, line 295: Should airflow also be mapped to airflow2, as it is the current version of airflow. ?

sbchaos (approved):

arinda-arif (approved):
LGTM

cmd/server/server.go
sravankorumilli: agree on this, making the change

PR #221, opened 2022-03-14
refactor: fix linter for init and autofix gofumpt

sbchaos:


.golangci.yml
```diff
@@ -58,18 +56,21 @@ linters:
...
   gomnd:
     ignored-numbers: # Why we have a big range of file permissions
-      - '0644'
-      - '0655'
-      - '0666'
-      - '0770'
-      - '0755'
-      - '0765'
-      - '0777'
+      - '0o644'
```
sravankorumilli, line 67: check on the impact of the converting to octal notation & doesn't seem to be intuitive for a user too

sbchaos: This is the issue from the go where gofmt talks about making the octal notion clear: https://github.com/golang/go/issues/33363

sbchaos: Golang version after 1.13 should prefer the octal notation to make it clear.
const perm = 0o755

sravankorumilli (approved):

Issue #180, opened 2022-02-25
Support Recompiling & Refreshing of Jobs in the scheduler.


sravankorumilli:
As per [RFC](https://github.com/odpf/optimus/pull/152) let's have a new command to refresh jobs such that 
they are recompiled, upgraded with the latest plugin.

**Scenarios**

1. Command should be able to refresh all or specific jobs of a project/namespace.
2. Dependencies should be persisted
3. Persisted dependencies should be used for job compilation & deployment.
4. Deploying the entire project instead of specific namespace which is happening as of now.
5. Job Deployment should not fail if any job compilation is failing it should proceed with the rest.

**Out of Scope**
1. Job Deployment as async operation

`optimus job refresh`

arinda-arif:
will introduce below rpc:
```proto
RefreshJobs(RefreshJobsRequest) returns (stream RefreshJobsResponse)

message RefreshJobsRequest {
	string project_name = 1;
	repeated NamespaceJobs namespace_jobs = 2;
}

message NamespaceJobs {
	string namespace_name = 1;
	repeated job_names = 2;
}

message RefreshJobsResponse {
	bool success = 1;

        // non ack responses are more of a progress/info response
        // and not really success or failure statuses
	bool ack = 2;

	string message = 3;
	string namespace_name = 4;
	string job_name = 5;
}
```

Note:
- For refresh, it is `RefreshJobs` not `RefreshJobSpecifications` as it is not requesting any job specs.
- Request is not a stream as we are not passing job specifications.
- Namespace is included as part of response as we are supporting multiple namespace job refresh.

PR #227, opened 2022-03-17
feat: add job refresh

arinda-arif:
https://github.com/odpf/optimus/issues/180

Acceptance criteria:
- Should able to refresh:
    - selected jobs, or 
    - all jobs in selected namespaces, or 
    - all jobs in a project
- After doing refresh, ALL jobs in the requested project will be deployed
- Dependency should be correctly persisted
- If dependency resolution is failed for a job, it should not failed the whole refresh process
- Users should know which jobs failed on refresh or deploy
- Metrics on every process in refresh should be pushed


arinda-arif:
https://github.com/odpf/proton/pull/106

store/postgres/migrations/000020_add_job_dependencies_table.down.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE job DROP IF EXISTS external_dependencies;
```
sbchaos, line 1: We should be dropping table `job_dependency`.

store/store.go
```diff
@@ -40,6 +43,7 @@ type ProjectJobSpecRepository interface {
 type ProjectRepository interface {
 	Save(context.Context, models.ProjectSpec) error
 	GetByName(context.Context, string) (models.ProjectSpec, error)
+	GetByID(context.Context, uuid.UUID) (models.ProjectSpec, error)
```
sbchaos, line 46: Please create a new model in Project for the `ProjectID` and use it in here.

store/store.go
```diff
@@ -115,3 +119,10 @@ type BackupRepository interface {
 	GetAll(context.Context) ([]models.BackupSpec, error)
 	GetByID(context.Context, uuid.UUID) (models.BackupSpec, error)
 }
+
+// JobDependencyRepository represents a storage interface for job dependencies
+type JobDependencyRepository interface {
+	Save(ctx context.Context, projectID, jobID uuid.UUID, dependency models.JobSpecDependency) error
```
sbchaos, line 125: Use model for `ProjectID` here, so that both do not get mixed.

store/postgres/migrations/000020_add_job_dependencies_table.up.sql
```diff
@@ -0,0 +1,9 @@
+CREATE EXTENSION IF NOT EXISTS "uuid-ossp";
+CREATE TABLE IF NOT EXISTS job_dependency (
+    job_id UUID NOT NULL,
+    project_id UUID NOT NULL,
```
sbchaos, line 4: Should we create a ForeignKey constraint for `project_id`.

store/postgres/job_spec_repository.go
```diff
@@ -63,6 +63,27 @@ func (repo *ProjectJobSpecRepository) GetByName(ctx context.Context, name string
 	return jobSpec, namespaceSpec, nil
 }
 
+func (repo *ProjectJobSpecRepository) GetByIDs(ctx context.Context, jobIDs []uuid.UUID) ([]models.JobSpec, error) {
+	var jobs []Job
+	if err := repo.db.WithContext(ctx).Where("project_id = ? AND job.id in ?", jobIDs).First(&jobs).Error; err != nil {
```
sbchaos, line 68: Are we able to query multiple jobs using `First` ?

As per gorm docs:
GORM provides First, Take, Last methods to retrieve a single object from the database, it adds LIMIT 1 condition when querying the database

store/postgres/job_dependency_repository_test.go
```diff
@@ -0,0 +1,93 @@
...
+	"os"
+	"testing"
+
+	"github.com/google/uuid"
+	"github.com/odpf/optimus/models"
+	"github.com/stretchr/testify/assert"
+	"gorm.io/gorm"
+)
+
+func TestJobDependencyRepository(t *testing.T) {
```
sbchaos, line 17: The name of function should be `TestIntegrationJobDependencyRepository`, we use name TestIntegration to identify the integration tests.

store/postgres/job_dependency_repository_test.go
```diff
@@ -0,0 +1,93 @@
...
+		m, err := NewHTTPFSMigrator(dbURL)
+		if err != nil {
+			panic(err)
+		}
+		if err := m.Drop(); err != nil {
+			panic(err)
+		}
+		if err := Migrate(dbURL); err != nil {
+			panic(err)
+		}
```
sbchaos, line 46: This should not be required, now we run the migrations once before any test is run.
Please use following in place of code above
```
		dbConn := setupDB()
		truncateTables(dbConn)
```

store/postgres/job_dependency_repository_test.go
```diff
@@ -0,0 +1,93 @@
...
+
+		projRepo := NewProjectRepository(dbConn, hash)
+		assert.Nil(t, projRepo.Save(ctx, projectSpec))
+		return dbConn
+	}
+
+	t.Run("Save", func(t *testing.T) {
+		db := DBSetup()
+		sqlDB, _ := db.DB()
+		defer sqlDB.Close()
```
sbchaos, line 56: These 2 lines should not be required after the above change.

store/postgres/job_dependency_repository_test.go
```diff
@@ -0,0 +1,93 @@
...
+
+		jobID1 := uuid.Must(uuid.NewRandom())
+		jobID2 := uuid.Must(uuid.NewRandom())
+		jobID3 := uuid.Must(uuid.NewRandom())
```
sbchaos, line 60: Please use `uuid.New()` which is same thing.

store/postgres/job_spec_repository.go
```diff
@@ -63,6 +63,27 @@ func (repo *ProjectJobSpecRepository) GetByName(ctx context.Context, name string
 	return jobSpec, namespaceSpec, nil
 }
 
+func (repo *ProjectJobSpecRepository) GetByIDs(ctx context.Context, jobIDs []uuid.UUID) ([]models.JobSpec, error) {
```
sbchaos, line 66: Please add test for this method.

store/postgres/job_dependency_repository.go
```diff
@@ -0,0 +1,80 @@
...
+	DependentProject   Project   `gorm:"foreignKey:DependentProjectID"`
+
+	Type string
+
+	CreatedAt time.Time `gorm:"not null" json:"created_at"`
+}
+
+type jobDependencyRepository struct {
+	db      *gorm.DB
+	project models.ProjectSpec
```
sbchaos, line 28: Please do not use project in repository, pass it as a method parameter.

cmd/server/server.go
```diff
@@ -178,6 +178,14 @@ func (fac *backupRepoFactory) New(projectSpec models.ProjectSpec, storer models.
 	return postgres.NewBackupRepository(fac.db, projectSpec, storer)
 }
 
+type jobDependencyRepoFactory struct {
```
sbchaos, line 181: Can we get rid of repoFactory and use the repo directly.

store/postgres/migrations/000020_add_job_dependencies_table.down.sql
arinda-arif: updated

store/postgres/migrations/000020_add_job_dependencies_table.up.sql
arinda-arif: added for both `project_id` and `dependent_project_id`. also have created index for `job_id` and `project_id`.

store/postgres/job_spec_repository.go
arinda-arif: you are right, instead of `First` this should use `Find`. updated.

store/postgres/job_dependency_repository_test.go
arinda-arif: updated

store/postgres/job_dependency_repository_test.go
arinda-arif: updated

store/postgres/job_dependency_repository_test.go
arinda-arif: removed

store/postgres/job_dependency_repository_test.go
arinda-arif: updated

store/postgres/job_spec_repository.go
arinda-arif: test added

cmd/server/server.go
arinda-arif: removed

store/postgres/job_dependency_repository_test.go
arinda-arif: also need to add to truncateTables list

store/store.go
arinda-arif: deleted `GetByID` from project repository as it is only being used inside `project_repository_test`. to verify the test, i have replaced `GetByID` with `GetByName`.

store/store.go
arinda-arif: have added models.ProjectID and use it in ProjectSpec

arinda-arif:
Hi @sbchaos, thank you for your feedback.

I have made the changes to resolve them. Please help to retake a look.

job/service.go
```diff
@@ -59,11 +82,16 @@ type ProjectRepoFactory interface {
 	New() store.ProjectRepository
 }
 
+// DependencyRepoFactory is used to manage job dependencies from store
+type DependencyRepoFactory interface {
```
sbchaos, line 86: Can we please remove this

models/project.go
```diff
@@ -37,8 +37,14 @@ var PluginSecretString = func(pluginType InstanceType, pluginName string) string
 	return strings.ToUpper(fmt.Sprintf("%s_%s", pluginType, pluginName))
 }
 
+type ProjectID uuid.UUID
+
+func (p ProjectID) UUID() uuid.UUID {
```
sbchaos, line 42: 👍

sbchaos (approved):

job/service.go
arinda-arif: removed

job/service.go
```diff
@@ -738,3 +692,194 @@ func populateDownstreamDAGs(dagTree *tree.MultiRootTree, jobSpec models.JobSpec,
...
+	if err = srv.resolveAndPersistDependency(ctx, projectSpec, namespaceJobNamePairs, progressObserver); err != nil {
+		// if err is caused by dependency resolution, ignore this as error.
+		var merrs *multierror.Error
+		if !errors.As(err, &merrs) {
+			return err
+		}
+	}
+
+	// Fetch dependency and enrich
+	jobDependencies, err := srv.dependencyResolver.FetchJobDependencies(ctx, projectSpec, progressObserver)
```
sravankorumilli, line 710: Similarly here, as you are refetching the dependencies which were constructed above i believe your intention is to move to seperate place later. It would be better to do that right now.
1. FetchAllJobsAgain
2. FetchAllJobDependencies
3. EnrichJobsWithJobDependencies
4. EnrichJobsWithHookDependencies // This needs to be optimized by not making call to plugin inside enrichment rather we cache the dependencies between plugins during bootstrap and use pluginService for these kind of purposes.
5. PriorityResolution 
6. Grouping Namespecs // This step too, why are we keep on passing the projectJobSpecRep and fetching jobs lets just enrich the JobSpecModel with Project & NamespaceSpec or store ProjectId and NamespaceID in the JobSpec which can be used for grouping.

I expect all the data from datasources or third party integrations to happen only once and this data can be cached for optimization purposes. in here we are passing jobspecrepo in every place and fetching jobspecs again and again which can be avoided.

job/service.go
```diff
@@ -738,3 +692,194 @@ func populateDownstreamDAGs(dagTree *tree.MultiRootTree, jobSpec models.JobSpec,
...
+
+	// enrich hook dependencies
+	for _, jobSpec := range jobSpecs {
+		jobSpec.Hooks = srv.dependencyResolver.FetchHookWithDependencies(jobSpec)
+	}
+
+	return jobSpecs, nil
+}
+
+func (srv *Service) prepareJobSpecs(ctx context.Context, projectSpec models.ProjectSpec,
```
sravankorumilli, line 852: prepare is generic better to have specific functions so readers can easily understand whats happening

job/service.go
```diff
@@ -738,3 +692,194 @@ func populateDownstreamDAGs(dagTree *tree.MultiRootTree, jobSpec models.JobSpec,
...
+	// compile assets before resolving in parallel
+	for i, jSpec := range jobSpecs {
+		if jobSpecs[i].Assets, err = srv.assetCompiler(jSpec, srv.Now()); err != nil {
+			return fmt.Errorf("asset compilation: %w", err)
+		}
+	}
+
+	// resolve specs in parallel
+	runner := parallel.NewRunner(parallel.WithTicket(ConcurrentTicketPerSec), parallel.WithLimit(ConcurrentLimit))
+	for _, jobSpec := range jobSpecs {
```
sravankorumilli, line 769: If possible we can see a mechanism on abstract out all the operations of how the parallel running is happening by keeping things declarative

job/dependency_resolver.go
```diff
@@ -52,6 +55,131 @@ func (r *dependencyResolver) Resolve(ctx context.Context, projectSpec models.Pro
...
+		// insert the new ones
+		err := r.dependencyRepo.Save(ctx, projectSpec.ID, jobSpec.ID, dependency)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (r *dependencyResolver) FetchJobDependencies(ctx context.Context, projectSpec models.ProjectSpec,
```
sravankorumilli, line 98: FetchJobDependencies I believe can be simplified the logic of fetching and updating with inter/intra job dependencies .

sravankorumilli (changes requested):

job/service.go
```diff
@@ -25,13 +28,33 @@ const (
...
+		Help: "Duration of resolving job dependency",
+	})
+)
 
 type AssetCompiler func(jobSpec models.JobSpec, scheduledAt time.Time) (models.JobAssets, error)
 
 // DependencyResolver compiles static and runtime dependencies
 type DependencyResolver interface {
 	Resolve(ctx context.Context, projectSpec models.ProjectSpec, jobSpec models.JobSpec, observer progress.Observer) (models.JobSpec, error)
+	ResolveAndPersist(ctx context.Context, projectSpec models.ProjectSpec, jobSpec models.JobSpec, observer progress.Observer) error
```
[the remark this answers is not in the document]
arinda-arif: agree, I think it will reduce the complexity. however, in the existing resolve function, there is a `resolveHookDependencies` process, which is not needed (as we are going to store the requested and fetch all after). should we also decouple it from `Resolve` or keep it until we refactor job `Sync`?

job/service.go
```diff
@@ -738,3 +692,194 @@ func populateDownstreamDAGs(dagTree *tree.MultiRootTree, jobSpec models.JobSpec,
 
 	return rootNode, nil
 }
+
+// Refresh fetches all the requested jobs, resolves its dependencies, assign proper priority weights,
+// compile all jobs in the project and upload them to the destination store.
+func (srv *Service) Refresh(ctx context.Context, projectSpec models.ProjectSpec, namespaceJobNamePairs []models.NamespaceJobNamePair,
```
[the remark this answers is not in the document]
arinda-arif: I was limiting the usage of protobuf model to only in the handler and not passing it to the service layer. this `namespaceJobNamePairs ` is only being used to get all of the jobSpecs for dependency resolution. should we do it early in the handler? what do you think?

job/service.go
```diff
@@ -738,3 +692,194 @@ func populateDownstreamDAGs(dagTree *tree.MultiRootTree, jobSpec models.JobSpec,
...
+	progressObserver progress.Observer) (err error) {
+	// Resolve dependency
+	if err = srv.resolveAndPersistDependency(ctx, projectSpec, namespaceJobNamePairs, progressObserver); err != nil {
+		// if err is caused by dependency resolution, ignore this as error.
+		var merrs *multierror.Error
```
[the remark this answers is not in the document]
arinda-arif: there are several possible errors that happens in that particular function: multierror (errors from dependency resolution), when compiling assets, or when fetching the job specs. if it is from dependency resolution, we should skip the error, as we don't want the whole refresh process to be canceled. looks like it is not clear enough, i will refactor it.

job/service.go
```diff
@@ -738,3 +692,194 @@ func populateDownstreamDAGs(dagTree *tree.MultiRootTree, jobSpec models.JobSpec,
 
 	return rootNode, nil
 }
+
+// Refresh fetches all the requested jobs, resolves its dependencies, assign proper priority weights,
+// compile all jobs in the project and upload them to the destination store.
+func (srv *Service) Refresh(ctx context.Context, projectSpec models.ProjectSpec, namespaceJobNamePairs []models.NamespaceJobNamePair,
+	progressObserver progress.Observer) (err error) {
+	// Resolve dependency
+	if err = srv.resolveAndPersistDependency(ctx, projectSpec, namespaceJobNamePairs, progressObserver); err != nil {
```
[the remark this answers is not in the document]
arinda-arif: agree, will change.

job/service.go
```diff
@@ -25,13 +28,33 @@ const (
 	ConcurrentLimit        = 600
 )
 
-var errDependencyResolution = fmt.Errorf("dependency resolution")
+var (
+	errDependencyResolution = fmt.Errorf("dependency resolution")
+
+	resolveDependencyFailureGauge = promauto.NewGauge(prometheus.GaugeOpts{
```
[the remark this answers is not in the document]
arinda-arif: have changed it to only a single gauge. differentiate the metric values with `status` label.

job/service.go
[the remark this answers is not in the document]
arinda-arif: As discussed, reuse resolve and introduce `Persist` instead of `ResolveAndPersist`

job/service.go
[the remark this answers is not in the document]
arinda-arif: As discussed, still use `NamespaceJobNamePair` to avoid using external contract in the service layer. However, we are not passing NamespaceSpec, but NamespaceName instead (and fetching the spec just before where it is being used).

job/service.go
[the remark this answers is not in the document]
arinda-arif: Have modified this so it will not check for multierror, it will check for the error type.

job/service.go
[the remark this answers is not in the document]
arinda-arif: modifiedthis, however persist dependencies and resolve dependencies are still in a same parallel runner. please check this.

api/handler/v1beta1/job_spec.go
```diff
@@ -285,11 +286,58 @@ func (sv *JobSpecServiceServer) DeleteJobSpecification(ctx context.Context, req
...
+
+	observers := new(progress.ObserverChain)
+	observers.Join(sv.progressObserver)
+	observers.Join(&jobRefreshObserver{
+		stream: respStream,
+		log:    sv.l,
+		mu:     new(sync.Mutex),
+	})
+
+	namespaceJobNamePairs, err := sv.prepareNamespaceJobNamePair(respStream.Context(), req)
```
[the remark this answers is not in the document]
arinda-arif: as discussed, we will still use namespaceJobNamePairs to avoid using external contract in service layer.

models/job.go
```diff
@@ -411,3 +413,10 @@ type JobQuery struct {
 	Filter      []string
 	OnlyLastRun bool
 }
+
+type JobIDDependenciesPair struct {
+	JobID            uuid.UUID
+	DependentProject ProjectSpec
+	DependentJobID   uuid.UUID
+	Type             JobSpecDependencyType
```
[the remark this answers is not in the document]
arinda-arif: The values of JobSpecDependencyType are in the same file, in the const section (top of the file). This should be enough?

job/dependency_resolver.go
```diff
@@ -52,6 +55,131 @@ func (r *dependencyResolver) Resolve(ctx context.Context, projectSpec models.Pro
...
+	observer progress.Observer) (map[uuid.UUID][]models.JobSpecDependency, error) {
```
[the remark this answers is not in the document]
arinda-arif: As discussed, we are still using uuid for now, to avoid having very big change in 1 PR

job/service.go
arinda-arif: updated. please help to recheck

job/service.go
arinda-arif: renamed this to `fetchJobSpecs`

job/dependency_resolver.go
arinda-arif: have modified this, FetchJobDependencies will just fetch the dependencies, adapting to the required model is handled on the enriching part (in deployer).

job/service.go
arinda-arif: decided to not refactor this part in this PR, as abstracting this out is quite tricky, and will see a better way to do this.

cmd/job_refresh.go
```diff
@@ -0,0 +1,177 @@
...
+			break
+		}
+
+		switch resp.Type {
+		case models.ProgressTypeJobUpload:
+			deployCounter++
+			if !resp.GetSuccess() {
+				deployFailedCounter++
+				if verbose {
+					l.Info(coloredError(fmt.Sprintf("%d. %s failed to be deployed: %s", deployCounter, resp.GetJobName(), resp.GetMessage())))
```
sravankorumilli, line 127: errors can be warn messages

cmd/job_refresh.go
```diff
@@ -0,0 +1,177 @@
...
+				if verbose {
+					l.Info(fmt.Sprintf("%d. %s successfully deployed", deployCounter, resp.GetJobName()))
+				}
+			}
+		case models.ProgressTypeJobDependencyResolution:
+			refreshCounter++
+			if !resp.GetSuccess() {
+				refreshFailedCounter++
+				if verbose {
+					l.Info(coloredError(fmt.Sprintf("error '%s': failed to refresh dependency, %s", resp.GetJobName(), resp.GetMessage())))
```
sravankorumilli, line 141: failures to be considered as warn logs

cmd/job_refresh.go
```diff
@@ -0,0 +1,177 @@
...
+			JobNames:      jobs,
+		})
+	}
+
+	jobSpecService := pb.NewJobSpecificationServiceClient(conn)
+	respStream, err := jobSpecService.RefreshJobs(refreshTimeoutCtx, &pb.RefreshJobsRequest{
+		ProjectName:   projectName,
+		NamespaceJobs: namespaceJobs,
+	})
+	if err != nil {
```
sravankorumilli, line 97: this nested if can be merged by just using single if on errors.Is

api/handler/v1beta1/job_spec.go
```diff
@@ -285,11 +286,45 @@ func (sv *JobSpecServiceServer) DeleteJobSpecification(ctx context.Context, req
...
+	observers.Join(sv.progressObserver)
+	observers.Join(&jobRefreshObserver{
+		stream: respStream,
+		log:    sv.l,
+		mu:     new(sync.Mutex),
+	})
+
+	namespaceJobNamePairs := sv.prepareNamespaceJobNamePairs(req.NamespaceJobs)
+
+	if err := sv.jobSvc.Refresh(respStream.Context(), req.ProjectName, namespaceJobNamePairs, observers); err != nil {
```
sravankorumilli, line 302: can you split this into multiple lines, will be more readable

models/namespace.go
```diff
@@ -16,3 +16,8 @@ type NamespaceSpec struct {
 }
 
 const AllNamespace = "*"
+
+type NamespaceJobNamePair struct {
```
sravankorumilli, line 20: we don't need this struct, we can avoid this.

api/handler/v1beta1/job_spec.go
sravankorumilli: Can you change the proto request to accept request for project refresh or bunch of namespaces in project or bunch of jobs. project is mandatory. namespaces and jobs list is optional.
We can have different functions for refreshing all jobs for a project. refresh all jobs of given namespaces, refresh bunch of jobs in a project.

job/service.go
```diff
@@ -738,3 +705,120 @@ func populateDownstreamDAGs(dagTree *tree.MultiRootTree, jobSpec models.JobSpec,
...
+			if errors.Is(err, errAssetCompilation) {
+				return err
+			}
+		}
+	}
+
+	return srv.deployer.Deploy(ctx, projectSpec, progressObserver)
+}
+
+func (srv *Service) fetchJobSpecs(ctx context.Context, projectSpec models.ProjectSpec,
```
sravankorumilli, line 737: fetchJobSpecs should be split into 4 functions, fetchAllForAProject, fetchAllForGivenNamespaces, fetchSpecsForGivenJobNames,  fetchJobSpecs will be the wrapper function

sravankorumilli (changes requested):

api/handler/v1beta1/job_spec.go
arinda-arif: is this what you meant?
```proto
message RefreshJobsRequest {
  string project_name = 1;
  repeated string namespace_names = 2;
  repeated string job_names = 3;
}
```
currently, to refresh a bunch of jobs in a project, we need namespace info of where the jobs belong (for auth). we can go with what you proposed, but will still require the namespace that needs to be provided in the `namespace_names` (which is not straightforward), or we can avoid requesting the namespace for this case.

cmd/job_refresh.go
arinda-arif: the reason behind having nested if in error is to give more information for the specific error (in this case time out), besides showing a message of other errors. what do you think?

cmd/job_refresh.go
```diff
@@ -0,0 +1,177 @@
...
+	var refreshErrors []string
+	refreshCounter, refreshSuccessCounter, refreshFailedCounter := 0, 0, 0
+
+	var deployErrors []string
+	deployCounter, deploySuccessCounter, deployFailedCounter := 0, 0, 0
+
+	var streamError error
+	for {
+		resp, err := respStream.Recv()
+		if err != nil {
```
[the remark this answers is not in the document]
arinda-arif: the intention of having nested ifs is to ignore EOF and throw other errors. another option is to not make nested, but multiple ifs that are quite similar.

job/deployer.go
```diff
@@ -0,0 +1,150 @@
...
+		}
+		if err := d.batchScheduler.DeployJobs(ctx, jobs[0].NamespaceSpec, jobs, progressObserver); err != nil {
+			deployError = multierror.Append(deployError, err)
+		}
+	}
+
+	return deployError
+}
+
+func (d *deployer) enrichJobSpecWithJobDependencies(ctx context.Context, jobSpecs []models.JobSpec, jobDependencies []models.JobIDDependenciesPair) ([]models.JobSpec, error) {
```
sravankorumilli, line 73: we can refactor this method
1. createJobSpecMap can be dissolved into fetchJobSpecsForDependentJobs which can go into the main function call and the map creation will just  create JobSpecMap given jobSpecs and dependentJobSpecs, here we can optimize by grouping the jobs by dependent project and fetching only once. Or why not fetch the jobspecs through a foreign key contraint, then we can avoid this fetch altogether.
2. We will have a function to enrich a single obSpec With JobDependencies, which will be reused here. When ever there is nesting better to break down into multiple functions

sravankorumilli (changes requested):

cmd/job_refresh.go
arinda-arif: updated

cmd/job_refresh.go
arinda-arif: updated

api/handler/v1beta1/job_spec.go
arinda-arif: updated, the proto PR: https://github.com/odpf/proton/pull/127

models/namespace.go
arinda-arif: removed this as we are no longer need namespace to refresh specific jobs.

job/service.go
```diff
@@ -738,3 +705,120 @@ func populateDownstreamDAGs(dagTree *tree.MultiRootTree, jobSpec models.JobSpec,
...
+	}
+}
+
+func (srv *Service) resolveDependency(ctx context.Context, projectSpec models.ProjectSpec,
+	jobSpecs []models.JobSpec, progressObserver progress.Observer) (err error) {
+	start := time.Now()
+	defer resolveDependencyHistogram.Observe(time.Since(start).Seconds())
+
+	// compile assets before resolving in parallel
+	for i, jSpec := range jobSpecs {
```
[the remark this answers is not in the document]
arinda-arif: updated

job/deployer.go
arinda-arif: updated as discussed: not fetching the jobspecs through foreign key constraint (to avoid performance issue), jobSpecMap and enrichment are now being done when fetching job specs with job dependencies.

job/service.go
```diff
@@ -738,3 +704,129 @@ func populateDownstreamDAGs(dagTree *tree.MultiRootTree, jobSpec models.JobSpec,
...
+	// get job specs as requested
+	jobSpecs, err := srv.fetchJobSpecs(ctx, projectSpec, namespaceNames, jobNames, progressObserver)
+	if err != nil {
+		return err
+	}
+
+	// resolve dependency and persist
+	if err := srv.resolveDependency(ctx, projectSpec, jobSpecs, progressObserver); err != nil {
+		// if err is caused by other than asset compilation, ignore the error.
+		if err != nil {
```
sravankorumilli, line 726: why fail in case of asset compilation

cmd/job_refresh.go
```diff
@@ -0,0 +1,167 @@
...
+		verbose     bool
+		cmd         = &cli.Command{
+			Use:     "refresh",
+			Short:   "Refresh job deployments",
+			Long:    "Redeploy jobs based on current server state",
+			Example: "optimus job refresh",
+		}
+	)
+
+	cmd.Flags().StringVarP(&projectName, "project", "p", projectName, "Optimus project name")
```
sravankorumilli, line 35: looks like we are accepting the flags but it doesn't work, as we are always referring from config. @deryrahman is working on fixing loading of specs from a single place, lets do the same here.

sravankorumilli (changes requested):

cmd/job_refresh.go
deryrahman: yes, it's known issue @sravankorumilli , will be fixed on https://github.com/odpf/optimus/pull/274, let's merge it once the PR is approved

job/service.go
arinda-arif: this should be removed by now. fixing

arinda-arif: removed

cmd/job_refresh.go
arinda-arif: for now i have removed the project flag

sravankorumilli (approved):

PR #228, opened 2022-03-17
[Draft] docs: add rfc for improving plugin ecosystem

sravankorumilli:


kushsharma:
I was also thinking in these terms, making plugin development as easy as possible is very critical for increasing optimus usage. Best alternate I think at the moment other than what we do is using web assembly. Sad part is, writing `wasm` in `go` and then running it in `go` is not matured yet(other than using `tinygo`) so we might look into either writing `wasm` plugins in `js` or `rust`.

ravisuhag:
@SaikumarChintada Can we please update this RFC as per final approach and close it?

SaikumarChintada:
@sravankorumilli can this PR be closed/archived in the interest of #356.

PR #230, opened 2022-03-21
refactor: enable gosec to check for security issues

sbchaos:


main.go
```diff
@@ -8,18 +8,16 @@ import (
 	"strings"
 	"time"
 
-	"github.com/sirupsen/logrus"
-
-	_ "net/http/pprof"
```
sbchaos, line 13: Exposes debug information as side-effect on the server port. This information already exposed on the profile address in telemetry.

utils/file.go
```diff
@@ -10,7 +10,7 @@ func WriteStringToFileIndexed() func(filePath, data string, writer io.Writer) er
 	index := 0
 	return func(filePath, data string, writer io.Writer) error {
 		if err := ioutil.WriteFile(filePath,
-			[]byte(data), 0644); err != nil {
+			[]byte(data), 0600); err != nil {
```
sbchaos, line 13: Check if it impacts upload to gcs bucket.

sravankorumilli (approved):

PR #240, opened 2022-03-24
refactor: break down server initialisation in different parts

sbchaos:


cmd/server/optimus.go
```diff
@@ -0,0 +1,347 @@
...
+		conf:       conf,
+		logger:     l,
+		serverAddr: addr,
+	}
+
+	if err := checkRequiredConfigs(conf.Server); err != nil {
+		return server, err
+	}
+
+	fns := []setupFn{
```
sravankorumilli, line 57: rename the fns as setupFNs and move startListening outside and call it explicitly

sravankorumilli (approved):

PR #242, opened 2022-03-25
fix: handle empty nameSpaces config to prevent infinite prompt loop

Mryashbhardwaj:
optimus falls into a infinite loop while asking to select for nameSpaces if nameSpaces are empty in the config file

### DIFF
--- a/cmd/namespace.go
+++ b/cmd/namespace.go
@@ -1,16 +1,21 @@
 package cmd
 
 import (
+	"errors"
+
 	"github.com/odpf/optimus/config"
 
 	"github.com/AlecAivazis/survey/v2"
 	"github.com/odpf/salt/log"
 )
 
-func askToSelectNamespace(l log.Logger, conf config.Optimus) *config.Namespace {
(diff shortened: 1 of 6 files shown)


cmd/namespace.go
```diff
@@ -1,26 +1,30 @@
...
 
-func askToSelectNamespace(l log.Logger, conf config.Optimus) *config.Namespace {
+func askToSelectNamespace(l log.Logger, conf config.Optimus) (*config.Namespace, error) {
 	options := make([]string, len(conf.Namespaces))
 	for i, namespace := range conf.Namespaces {
 		options[i] = namespace.Name
 	}
+	if len(options) == 0 {
+		return nil, errors.New("no namespace found in config file")
+	}
```
sbchaos, line 19: Can you please check the length of `conf.Namespaces` and return before the loop if the slice is empty.

Mryashbhardwaj: sure, done

sbchaos (approved):

Issue #171, opened 2022-02-25
Use progress bar around version call and not after it


sravankorumilli:

sbchaos:
In file: `cmd/version.go` we start the progress bar after making the call to server. It should be started before the call.

PR #243, opened 2022-03-25
fix: use of progress on version call

siddhanta-rath:
closes https://github.com/odpf/optimus/issues/171

### DIFF
--- a/cmd/version.go
+++ b/cmd/version.go
@@ -78,15 +78,14 @@ func getVersionRequest(clientVer, host string) (ver string, err error) {
 	defer cancel()
 
 	runtime := pb.NewRuntimeServiceClient(conn)
+	spinner := NewProgressBar()
+	spinner.Start("please wait...")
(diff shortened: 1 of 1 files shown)


cmd/version.go
```diff
@@ -78,15 +78,14 @@ func getVersionRequest(clientVer, host string) (ver string, err error) {
...
 	versionResponse, err := runtime.Version(ctx, &pb.VersionRequest{
 		Client: clientVer,
 	})
 	if err != nil {
 		return "", fmt.Errorf("request failed for version: %w", err)
 	}
-
-	spinner := NewProgressBar()
-	spinner.Start("please wait...")
 	time.Sleep(versionTimeout)
```
sbchaos, line 89: Can you please remove the sleep call if not required.

siddhanta-rath: if I remove the `time.Sleep`, the spinner is not running

sbchaos (approved):

Issue #247, opened 2022-03-28
Update Job Status Command to use Job Run API


sravankorumilli:

siddhanta-rath:
changed existing job status command to :

- `optimus job runs sample_select` (default)
- `optimus job runs sample_select --start_date 2022-03-01T02:00:00Z --end_date 2022-03-12T00:00:00Z` (with date range)

PR #255, opened 2022-03-29
refactor: update job status command to use job run API

siddhanta-rath:
closes https://github.com/odpf/optimus/issues/247


cmd/job_run_list.go
```diff
@@ -0,0 +1,156 @@
...
+	req := &pb.JobRunRequest{
+		ProjectName: arg.projectName,
+		JobName:     arg.jobName,
+	}
+	return req, nil
+}
+
+func getJobRunRequestWithDate(arg jobRunCmdArg) (*pb.JobRunRequest, error) {
+	var req *pb.JobRunRequest
+	dateRange, err := validateDateRange(arg.startDate, arg.endDate)
```
sravankorumilli, line 124: why in validation you are returning a daterange and again just setting dateRange.start & dateRange.end validate can just return err and dateRange struct can be avoided altogether

cmd/job_run_list.go
```diff
@@ -0,0 +1,156 @@
...
+	var endDate string
+	cmd.Flags().StringVarP(&projectName, "project", "p", projectName, "Project name of optimus managed repository")
+	cmd.Flags().StringVar(&host, "host", defaultHost, "Optimus service endpoint url")
+	cmd.Flags().StringVar(&startDate, "start_date", "", "start date of job run")
+	cmd.Flags().StringVar(&endDate, "end_date", "", "end date of job run")
+	cmd.RunE = func(c *cli.Command, args []string) error {
+		jobName := args[0]
+		l.Info(fmt.Sprintf("Requesting status for project %s, job %s from %s",
+			projectName, jobName, host))
+		arg := jobRunCmdArg{
```
sravankorumilli, line 54: instead of creating the jobRunCmdArg and then the jobRunRequest and doing the validation somewhere down the line. Lets add just three functions.
1. One for validation.
2. One for requestCreation.
3. getJobRunList

We would not need anymore functions.

cmd/job_run_list.go
```diff
@@ -0,0 +1,156 @@
...
+	timeoutCtx, cancel := context.WithTimeout(context.Background(), jobStatusTimeout)
+	defer cancel()
+
+	run := pb.NewJobRunServiceClient(conn)
+	req, err := fn(arg)
+	if err != nil {
+		return err
+	}
+
+	spinner := NewProgressBar()
```
sravankorumilli, line 97: you can keep the spinner in the main command instead of job run api i believe spinner we start before we start processing a command and stops after all the processing is done so the main orchestrator of the functions can just take care of this.

cmd/job_run_list.go
```diff
@@ -0,0 +1,156 @@
...
+	if arg.startDate == "" && arg.endDate == "" {
+		return callJobRun(l, host, arg, getJobRunRequest)
+	}
+	// get runs between start_date and end_date (both start_date and end_date are inclusive)
+	return callJobRun(l, host, arg, getJobRunRequestWithDate)
+}
+
+func callJobRun(l log.Logger, host string, arg jobRunCmdArg, fn getRequestFn) error {
+	var err error
+	dialTimeoutCtx, dialCancel := context.WithTimeout(context.Background(), OptimusDialTimeout)
```
sravankorumilli, line 76: isn't it a best practice to create the context in the command and pass it to the needed functions from there instead of creating the context in the sub functions.

cmd/job_run_list.go
```diff
@@ -0,0 +1,156 @@
...
+	req = &pb.JobRunRequest{
+		ProjectName: arg.projectName,
+		JobName:     arg.jobName,
+		StartDate:   timestamppb.New(dateRange.start),
+		EndDate:     timestamppb.New(dateRange.end),
+	}
+	return req, nil
+}
+
+func validateDateRange(startDate, endDate string) (dateRange, error) {
```
sravankorumilli, line 137: validate shouldn't return anything other than error

cmd/job_run_list.go
```diff
@@ -0,0 +1,156 @@
...
+	start time.Time
+	end   time.Time
+}
+
+type getRequestFn func(jobRunCmdArg) (*pb.JobRunRequest, error)
+
+func jobRunListCommand(l log.Logger, defaultProjectName, defaultHost string) *cli.Command {
+	cmd := &cli.Command{
+		Use:     "runs",
+		Short:   "Get current job runs",
```
sravankorumilli, line 38: How about we have command  `optimus job list-runs` it is more clear I believe
Short: Get Job run details

cmd/job.go
```diff
@@ -22,6 +22,6 @@ func jobCommand(l log.Logger, conf config.Optimus, pluginRepo models.PluginRepos
 	cmd.AddCommand(jobRenderTemplateCommand(l, conf, pluginRepo))
 	cmd.AddCommand(jobValidateCommand(l, conf, pluginRepo, conf.Project.Name, conf.Host))
 	cmd.AddCommand(jobRunCommand(l, conf, pluginRepo, conf.Project.Name, conf.Host))
```
sravankorumilli, line 24: I don't think we use this command, can you remove the code for this command too, it will be little confusing

sravankorumilli (changes requested):

cmd/job_run_list.go
siddhanta-rath: done as part of latest commit

cmd/job_run_list.go
siddhanta-rath: addressed as part of latest commit

cmd/job_run_list.go
siddhanta-rath: context is passed from `getJobRunList` now

cmd/job_run_list.go
siddhanta-rath: 1. removed `jobRunCmdArg`
2. refactor `getJobRunList` to validate the cmd args, creating request and getting the response.
3. removed `getRequestFn` type 
4. removed `getJobRunRequest` function
5. removed `getJobRunRequestWithDate` function
6. simplify validate args function and renamed to `validateDateArgs `
7. handled request creation under single function `createJobRunRequest`

cmd/job_run_list.go
siddhanta-rath: handled as part of latest commit

cmd/job_run_list.go
```diff
@@ -0,0 +1,133 @@
...
+	req, err = createJobRunRequest(projectName, jobName, startDate, endDate)
+	if err != nil {
+		return err
+	}
+	ctx := context.Background()
+	err = callJobRun(l, host, ctx, req)
+	return err
+}
+
+func callJobRun(l log.Logger, host string, ctx context.Context, jobRunRequest *pb.JobRunRequest) error {
```
sravankorumilli, line 63: ctx to be the first argument

siddhanta-rath: done

sravankorumilli (approved):

PR #269, opened 2022-04-05
Refactor deploy

irainia:
This PR is to address the need to refactor `deploy` command.

To merge this PR, the following should be confirmed:

- [x] merge  #257

### DIFF
--- a/cmd/admin_build_instance.go
+++ b/cmd/admin_build_instance.go
@@ -1,8 +1,6 @@
 package cmd
 
 import (
-	"context"
-	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
@@ -11,7 +9,6 @@ import (
 
 	"github.com/odpf/salt/log"
 	cli "github.com/spf13/cobra"
-	"google.golang.org/grpc"
 	"google.golang.org/protobuf/types/known/timestamppb"
 
 	pb "github.com/odpf/optimus/api/proto/odpf/optimus/core/v1beta1"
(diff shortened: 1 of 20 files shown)


cmd/commands.go
```diff
@@ -133,6 +134,63 @@ func New() *cli.Command {
...
+	return map[string]string{
+		"Authorization": fmt.Sprintf("Basic %s", a.Token),
+	}, nil
+}
+
+func (a *BasicAuthentication) RequireTransportSecurity() bool {
+	return false
+}
+
+func initClientConnection(l log.Logger, serverHost string, requestTimeout time.Duration) (
```
siddhanta-rath, line 165: will it be a good idea to move all GRPC client-related code to a separate file?

irainia: yeah, that could be a good idea. will try to check it and update it accordingly.

cmd/admin_build_instance.go
```diff
@@ -81,24 +78,15 @@ func getInstanceBuildRequest(l log.Logger, jobName, inputDirectory, host, projec
...
 
-	dialTimeoutCtx, dialCancel := context.WithTimeout(context.Background(), OptimusDialTimeout)
-	defer dialCancel()
-
-	var conn *grpc.ClientConn
-	if conn, err = createConnection(dialTimeoutCtx, host); err != nil {
-		if errors.Is(err, context.DeadlineExceeded) {
-			l.Error(ErrServerNotReachable(host).Error())
-		}
+	ctx, conn, closeConn, err := initClientConnection(l, host, adminBuildInstanceTimeout)
```
sbchaos, line 81: Can you try pushing creation of dial context to createConnection. That context is only required at time of creating connection.

Also if err is context.DeadlineExceeded then return `ErrServerNotReachable` error.

cmd/commands.go
irainia: it's addressed

cmd/admin_build_instance.go
irainia: concerns are received from separate call, will check and update accordingly

irainia: concerns are addressed and some bugs are fixed.

cmd/connectivity.go
```diff
@@ -0,0 +1,142 @@
...
+) {
+	connection, err = createConnection(serverHost)
+	if err != nil {
+		return
+	}
+	reqCtx, reqCancel := context.WithTimeout(context.Background(), requestTimeout)
+
+	requestCtx = reqCtx
+	closeConnection = func() {
+		l.Info("Closing client connection")
```
sbchaos, line 91: Is this for testing or do we need this info log ?

sbchaos, line 91: Is this for testing or do we need this info log ?

sbchaos (approved):

cmd/connectivity.go
irainia: oh right, missed that one. should've removed it

irainia: it is addressed

PR #271, opened 2022-04-05
fix: prerun client config load

deryrahman:
- [x] proper load client config inside persistentPreRun
- [x] initial variable taken from conf instead of args (will fix on separate PR after cmd client refactoring done)

### DIFF
--- a/cmd/admin.go
+++ b/cmd/admin.go
@@ -1,7 +1,6 @@
 package cmd
 
 import (
-	"github.com/odpf/salt/log"
 	cli "github.com/spf13/cobra"
 
 	"github.com/odpf/optimus/config"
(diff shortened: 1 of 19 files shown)


arinda-arif:
@deryrahman LGTM. However for jobs, replay, backup commands would prefer the logger to be initialized in the `PersistentPreRunE`. It can be done in separate PR.

irainia:
For the most part, I think it's good and we can merge it.

deryrahman:
> @deryrahman LGTM. However for jobs, replay, backup commands would prefer the logger to be initialized in the `PersistentPreRunE`. It can be done in separate PR.

We're initializing logger inside command to minimize the parameter passing. We can do that as part of client command refactoring

PR #273, opened 2022-04-05
fix: improve logging for deployment

sbchaos:


cmd/deploy.go
```diff
@@ -263,12 +262,6 @@ func deployAllJobs(deployTimeoutCtx context.Context,
 		}
 	}
 	spinner.Stop()
-	if len(streamErrs) > 0 {
-		for _, e := range streamErrs {
-			l.Error(e.Error())
-		}
-		return errors.New("one or more errors are encountered during job deployment")
-	}
```
sbchaos, line 271: Unreachable code, if there is any error, this does not get executed and errors are not printed

sravankorumilli (approved):

PR #274, opened 2022-04-06
feat: bind config with cobra flags to override the conf

deryrahman:
- [x] overriding the config via flags provided by cobra
- [x] mapping the flag with `-` delimiter instead of `.` delimiter (--project-name instead of --project.name)
- [x] change flag names for each command that needs config overriding

TODO next (can be done in separate PR):
- [ ] bump salt version to support pflags
- [ ] rename `--project` to `--project-name` on entrypoint.sh for each plugins

### DIFF
--- a/cmd/admin_build_instance.go
+++ b/cmd/admin_build_instance.go
@@ -29,8 +29,6 @@ const unsubstitutedValue = "<no value>"
 
 func adminBuildInstanceCommand(conf *config.ClientConfig) *cli.Command {
 	var (
-		optimusHost    string
-		projectName    string
 		assetOutputDir = "/tmp/"
 		runType        = "task"
 		runName        string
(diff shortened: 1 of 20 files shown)


deryrahman:
@sravankorumilli by default, [salt is flattening the config keys with dot](https://github.com/odpf/salt/blob/60f6178cfd80cc80b4905e940a435501acb31125/config/config.go#L180) `.` delimiter, eg in server config (`config.sample.yaml`), we will have `serve.port` `serve.host` etc as their flags naming. eg. `./optimus serve --serve.port=8000`

Instead of dot `.`, it's better to have dash `-` delimiter, so it will be `./optimus serve --serve-port=8000`

I’m proposing the changes for salt config to have ability to replace the flag name delimiter, like we have on [EnvKeyReplacer](https://github.com/odpf/salt/blob/60f6178cfd80cc80b4905e940a435501acb31125/config/config.go#L170)

wdyt?

deryrahman:
fix https://github.com/odpf/optimus/pull/257 and https://github.com/odpf/optimus/pull/271

cmd/admin_build_instance.go
```diff
@@ -54,12 +52,12 @@ func adminBuildInstanceCommand(conf *config.ClientConfig) *cli.Command {
 	cmd.Flags().StringVar(&runName, "name", "", "Name of running instance, e.g., bq2bq/transporter/predator")
 	cmd.MarkFlagRequired("name")
 
-	cmd.Flags().StringVarP(&projectName, "project", "p", projectName, "Name of the optimus project") // TODO: fix overriding conf via args
-	cmd.Flags().StringVar(&optimusHost, "host", optimusHost, "Optimus service endpoint url")         // TODO: fix overriding conf via args
+	cmd.Flags().StringP("project-name", "p", defaultProjectName, "Name of the optimus project")
```
sravankorumilli, line 55: this change needs a corresponding change in the entrypoint.sh where ever it is referred when the optimus binary is upgraded in the plugins.

sravankorumilli (commented):
What is the scope of this PR? Is it handled everywhere?

deryrahman:
> What is the scope of this PR? Is it handled everywhere?

All command that use flags to override the config should be impacted

cmd/admin_build_instance.go
deryrahman: noted, after we bump the optimus version, we can later update the flag name in `endpoint.sh` for each plugins

sravankorumilli (approved):

deryrahman:
summarizing our discussions:

```
1. need to define which configs that are able to override and which are not
2. `--project` means to select the project, not to override the configs. Treat it as `kubectx`
3. action items: 
    - bump the plugins with `--project-name`
    - revisit it to provide `--project` for backward compatible or not
```

@ravisuhag @sravankorumilli

PR #279, opened 2022-04-08
Python tests ci pipeline

Mryashbhardwaj:


ext/scheduler/airflow2/tests/Makefile
```diff
@@ -1,4 +1,5 @@
 all: build
 
 test:
-	@python -m unittest discover .
```
sravankorumilli, line 4: Lets move this target into the base one as it becomes hard for discoverability & no one runs these tests

.github/workflows/scheduler-resources-test.yml
```diff
@@ -0,0 +1,20 @@
+name: test optimus resources
```
sravankorumilli, line 1: why not integrate in make test-ci target, and we don't need another workflow for running unit tests.

sravankorumilli (changes requested):

.github/workflows/scheduler-resources-test.yml
Mryashbhardwaj: sure

Mryashbhardwaj: sure

PR #281, opened 2022-04-08
refactor: remove unused method receivers

sbchaos:


sravankorumilli:
is this needed?

sbchaos:
This is to make code more readable, looking at the name we can know if the instance variable is used or not

PR #285, opened 2022-04-12
docs: add documentation for using secrets in job spec

sbchaos:


sravankorumilli (changes requested):
Lets remove the old approach of using secrets

PR #294, opened 2022-04-18
refactor: decouple postgres adapter from plugin service

Mryashbhardwaj:


job/service.go
```diff
@@ -125,7 +125,12 @@ type Service struct {
 // Create constructs a Job for a namespace and commits it to the store
 func (srv *Service) Create(ctx context.Context, namespace models.NamespaceSpec, spec models.JobSpec) error {
 	jobRepo := srv.jobSpecRepoFactory.New(namespace)
-	if err := jobRepo.Save(ctx, spec); err != nil {
+	jobDestinationResponse, err := srv.pluginService.GenerateDestination(ctx, spec, namespace)
+	if err != nil {
```
sbchaos, line 129: Do we need to check for dependency mod error here ?

service/job_run_service.go
```diff
@@ -63,7 +63,12 @@ func (s *jobRunService) GetScheduledRun(ctx context.Context, namespace models.Na
 			// more than once.
 			newJobRun.ExecutedAt = jobRun.ExecutedAt
 		}
-		if err := repo.Save(ctx, namespace, newJobRun); err != nil {
+		jobDestinationResponse, err := s.pluginService.GenerateDestination(ctx, jobSpec, namespace)
+		if err != nil {
```
sbchaos, line 67: Same: do we check for dependency mod error ?

ext/scheduler/prime/scheduler.go
```diff
@@ -47,8 +47,9 @@ func (s *Scheduler) DeployJobs(ctx context.Context, namespace models.NamespaceSp
 	}
 
 	repo := s.jobRunRepoFac.New()
+	jobDestination := "implement me" // fetch job destination from plugin service
```
sbchaos, line 50: Make this empty string, other flows also return empty string when no destination

sbchaos (approved):

PR #295, opened 2022-04-18
fix: avoid modifying the origional confs

sbchaos:


service/plugin_service.go
```diff
@@ -93,13 +93,17 @@ func (d DependencyPluginService) compileConfig(ctx context.Context, configs mode
 		compiler.From(secrets.ToMap()).WithName("secret"),
 	)
 
-	for i, config := range configs {
+	var confs models.PluginConfigs
+	for _, config := range configs {
 		compiledCnf, err := d.engine.CompileString(config.Value, tmplCtx)
 		if err == nil {
```
sravankorumilli, line 99: why are we ignoring the error, either we should fail or we need to add the config before compilation, in this situatin I believe we can throw an error

sbchaos: We will add the config before the error

PR #298, opened 2022-04-19
fix: use correct script to drop job_dependency on migration down

arinda-arif:
Migration 000020 down was wrongly set. It should drop the `job_dependency` table, not drop `job_dependency` field in the `job` table.

Note: `job_dependency` in `job` is not exist.

### DIFF
--- a/store/postgres/migrations/000020_add_job_dependencies_table.down.sql
+++ b/store/postgres/migrations/000020_add_job_dependencies_table.down.sql
@@ -1,4 +1,1 @@
 DROP TABLE IF EXISTS job_dependency;
-
-DROP INDEX IF EXISTS job_dependency_job_id_idx;
-DROP INDEX IF EXISTS job_dependency_project_id_idx;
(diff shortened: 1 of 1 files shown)


store/postgres/migrations/000020_add_job_dependencies_table.down.sql
```diff
@@ -1 +1,4 @@
-ALTER TABLE job DROP IF EXISTS job_dependency;
\ No newline at end of file
+DROP TABLE IF EXISTS job_dependency;
```
irainia, line 1: why do we drop the table first mbak instead of dropping the index first?

sbchaos (approved):

store/postgres/migrations/000020_add_job_dependencies_table.down.sql
arinda-arif: Turns out, dropping table also includes dropping the indexes ([ref](https://www.postgresql.org/docs/14/sql-droptable.html)). So, I will remove the drop index instead. @irainia @sbchaos

irainia: cool, that also works

irainia (approved):

PR #299, opened 2022-04-19
feat: on-demand job execution via prime executor

kushsharma:
Target `v1` for issue https://github.com/odpf/optimus/issues/300

// TODO: will add more details


sravankorumilli:
@kushsharma lets resume the discussion on this, as its been a while the effort was put on this. It is better to revisit the problem and the approach, would suggest to create an issue and start the discussion on it.

sravankorumilli:
We can close this PR and restart things as part of this https://github.com/odpf/optimus/issues/300

PR #304, opened 2022-04-20
build: donot override the latest optimus image on releases

sravankorumilli:


kushsharma:
Does this mean we are not releasing `latest` tag?

PR #311, opened 2022-04-25
docs: add replay rate limit rfc

arinda-arif:


kushsharma:
I understand 3rd approach should be feasible to work out but why do we need these extra configs of `EXEUCTION_PROJECT` and `REPLAY_EXECUTION_PROJECT`?

Also lets add this in RFC, which is the recommended approach so far.

arinda-arif:
@kushsharma Agree on the segregation of configs in various places. Will revisit this RFC. We can also reconsider other approaches such as creating a custom API for triggering if possible.

arinda-arif:
> > Optimus will check if there is a replay request for the particular job and date, if yes, `EXECUTION_PROJECT` will be replaced with `REPLAY_EXECUTION_PROJECT`.
> 
> @arinda-arif can you also help me understand how are we using these two new configs?

Hi @kushsharma , `EXECUTION_PROJECT` is not really a new config, it is the config that is being used by the bq2bq plugin to run their BQ job on. Although up until now, Optimus does not know this config exists, as it is being passed to the plugin and is used in the plugin. 

The point of having `REPLAY_EXECUTION_PROJECT` is basically the same but only being used for Replay requests. We can have a check inside Optimus and mark if the particular request comes from Replay. Based on that, the plugin will use the execution project based on the `REPLAY_EXECUTION_PROJECT` value.

Issue #264, opened 2022-04-05
Optimus Commands functions are all package scoped, it is better to group them under specific group command structs


sravankorumilli:
The current `cmd` package contains many implementations in the scope of accepting the user input on how to execute Optimus. There are some issues being observed for the current approach:

* many functionalities are defined within one package, making the package itself seems bigger with a lot of functions, variables, and constants that are accessible accross the package
* with many components defined within one package, some of it could conflict with one another, and there were cases during development that two components (in this example, variables) were defined but served the same purpose
* during development, IDE or text editor's suggestion could clutter with other functionalities

To address the mentioned issues above, one approach that can be done is by restructuring the package to grouping some similar functionalities, like for `job` command, it is put in one place like struct and/or package.

irainia:
The following is the approach internally discussed to address this issue. Basically, it is to group the related components into its own structure within a dedicated package.

* main.go

```go
package main

// ...

func main() {
	// ...
	root := cmd.NewRoot()
	if err := root.Execute(); err != nil {
		panic(err)
	}
	// ...
}
```

* cmd/root.go

```go
package cmd

type root struct {
	*cobra.Command
}

func NewRoot() *cobra.Command {
	output := &root{
		Command: &cobra.Command{
			// ...
		},
	}
	jobCommand := job.NewCommand()
	output.AddCommand(jobCommand)
	return output.Command
}
```

* cmd/job/job.go

```go
package cmd

type job struct {
	*cobra.Command
}

func NewCommand() *cobra.Command {
	output := &job{
		// ...
	}
	jobCreateCommand := NewCreateCommand("inside job create")
	output.AddCommand(jobCreateCommand)
	return output.Command
}
```

* cmd/job/job_create.go

```go
package cmd

type jobCreate struct {
	*cobra.Command

	message string
}

func (j *jobCreate) GetMessage() string {
	return fmt.Sprintf("message: %s", j.message)
}

func (j *jobCreate) Execute(cmd *cobra.Command, args []string) error {
	fmt.Println(j.GetMessage())
	return nil
}

func NewCreateCommand(msg string) *cobra.Command {
	output := &jobCreate{
		// ...
	}
	output.message = msg
	output.RunE = output.Execute
	return output.Command
}
```

irainia:
> The following is the approach internally discussed to address this issue. Basically, it is to group the related components into its own structure within a dedicated package.
> 
> * main.go
> 
> ```go
> package main
> 
> // ...
> 
> func main() {
[...]

PR #313, opened 2022-04-26
refactor: command

irainia:
This PR is to address #264 .

This PR is to refactor the currently existing `cmd` package. The following is the general changes in this PR:

* Each main command is put under a dedicated package. For example `job` package under `cmd` as `job` here is considered as the main command. Main command is the sub-command that is directly added by the root, mostly available under `cmd/commands.go`.
* Each sub-command of the main command is put in the same package. For example, `job create` is put under `job` package, but different file.
* Each main command and its sub-command (if exist) uses a dedicated structure if that command involves execution. For example, in `job` it has structure `jobCommand` with its own functionalities. But, if we check `admin build` sub command, there's no such structure as there's no business execution specific to that sub-command.
* Some refactors on some implementations, including renaming for better clarity or to simplification, extraction function, and more.
* Some bug fixes that was previously missed, such as namespace describe and listing secret.


cmd/admin/build_instance.go
```diff
@@ -0,0 +1,203 @@
...
+		ProjectName:  b.clientConfig.Project.Name,
+		JobName:      jobName,
+		ScheduledAt:  jobScheduledTimeProto,
+		InstanceType: pb.InstanceSpec_Type(pb.InstanceSpec_Type_value[utils.ToEnumProto(b.runType, "type")]),
+		InstanceName: b.runName,
+	}
+	return jobRun.RegisterInstance(conn.GetContext(), request)
+}
+
+func (b *buildInstanceCommand) getJobScheduledTimeProto() (*timestamppb.Timestamp, error) {
```
sbchaos, line 197: We are not changing the struct, we should not be using pointer unless struct is big.

cmd/admin/build_instance.go
```diff
@@ -0,0 +1,203 @@
...
+	for fileName, fileContent := range jobResponse.Context.Files {
+		filePath := filepath.Join(dirPath, fileName)
+		if err := writeToFileFn(filePath, fileContent, b.logger.Writer()); err != nil {
+			return fmt.Errorf("failed to write asset file at %s: %w", filePath, err)
+		}
+	}
+	return nil
+}
+
+func (b *buildInstanceCommand) sendInstanceRequest(jobName string, jobScheduledTimeProto *timestamppb.Timestamp) (*pb.RegisterInstanceResponse, error) {
```
sbchaos, line 178: Is the pointer required here?

cmd/admin/build_instance.go
```diff
@@ -0,0 +1,203 @@
...
+	if err := writeToFileFn(filePath, envFileBlob, b.logger.Writer()); err != nil {
+		return fmt.Errorf("failed to write asset file at %s: %w", filePath, err)
+	}
+	return nil
+}
+
+func (b *buildInstanceCommand) writeJobResponseMapToFiles(
+	jobResponse *pb.RegisterInstanceResponse, dirPath string,
+) error {
+	if err := os.MkdirAll(dirPath, 0o777); err != nil {
```
sbchaos, line 164: The file permission of `0o777` is too open, can we restrict it to `600` or similar

cmd/admin/build_instance.go
```diff
@@ -0,0 +1,203 @@
...
+
+	filePath := filepath.Join(dirPath, models.InstanceDataTypeEnvFileName)
+	writeToFileFn := utils.WriteStringToFileIndexed()
+	if err := writeToFileFn(filePath, envFileBlob, b.logger.Writer()); err != nil {
+		return fmt.Errorf("failed to write asset file at %s: %w", filePath, err)
+	}
+	return nil
+}
+
+func (b *buildInstanceCommand) writeJobResponseMapToFiles(
```
sbchaos, line 161: Can we rename the function to `writeJobAssetsToFiles`, current name is not very descriptive.

cmd/backup/create.go
```diff
@@ -0,0 +1,254 @@
...
+		if errors.Is(err, context.DeadlineExceeded) {
+			c.logger.Error(logger.ColoredError("Backup took too long, timing out"))
+		}
+		return fmt.Errorf("request failed to backup job %s: %w", backupRequest.ResourceName, err)
+	}
+	c.printBackupResponse(backupResponse)
+	return nil
+}
+
+func (c *createCommand) printBackupResponse(backupResponse *pb.CreateBackupResponse) {
```
sbchaos, line 135: Can we use non-pointer here

cmd/backup/create.go
```diff
@@ -0,0 +1,254 @@
...
+	c.logger.Info(logger.ColoredSuccess("Resource backup completed successfully:"))
+	for counter, result := range backupResponse.Urn {
+		c.logger.Info(fmt.Sprintf("%d. %s", counter+1, result))
+	}
+}
+
+func (c *createCommand) getAllowedDownstreamNamespaces(namespaceName string) []string {
```
sbchaos, line 142: Simlify to a function, pass allDownstream as param.
```go
func getAllowedDownstreamNamespaces(namespaceName string, allDownstream bool) []string {
  if allDownstream {
      return []string{"*"}
  }
  return []string{namespaceName}
}
```

cmd/version/version.go
```diff
@@ -0,0 +1,135 @@
...
+	v.printAllPluginInfos()
+	return nil
+}
+
+func (v *versionCommand) PostRunE(_ *cobra.Command, _ []string) error {
+	v.pluginCleanFn()
+	return nil
+}
+
+func (v *versionCommand) printAllPluginInfos() {
```
sbchaos, line 94: Check pointer semantic usage

cmd/replay/create.go
```diff
@@ -0,0 +1,295 @@
...
+
+	replayID, err := c.runReplayRequest(jobName, startDate, endDate)
+	if err != nil {
+		return err
+	}
+	c.logger.Info(logger.ColoredSuccess("Replay request created with id %s", replayID))
+	return nil
+}
+
+func (c *createCommand) getAllowedDownstreamNamespaces() []string {
```
sbchaos, line 124: Convert to function

cmd/replay/list.go
```diff
@@ -0,0 +1,104 @@
...
+		"Start Date",
+		"End Date",
+		"Ignore Downstream?",
+		"Requested At",
+		"Status",
+	})
+
+	for _, replaySpec := range replayListResponse.ReplayList {
+		table.Append([]string{
+			replaySpec.Id, replaySpec.JobName, replaySpec.StartDate.AsTime().Format(models.JobDatetimeLayout),
```
sbchaos, line 96: Please put single item on one line.

cmd/admin/build_instance.go
irainia: updated

cmd/admin/build_instance.go
irainia: updated

cmd/backup/create.go
irainia: updated

cmd/replay/create.go
irainia: updated

cmd/replay/list.go
irainia: updated

cmd/initialize/initialize.go
```diff
@@ -0,0 +1,115 @@
...
+func (i *initializeCommand) initClientConfig(clientConfig *config.ClientConfig) error {
+	if err := i.setupDirPathForClientConfig(clientConfig); err != nil {
+		return err
+	}
+	marshalledClientConfig, err := yaml.Marshal(clientConfig)
+	if err != nil {
+		return err
+	}
+	filePath := i.getClientConfigPath()
+	filePermission := 0o660
```
sbchaos, line 85: Do we need permission to be `0o660` ? Can we restrict it to `0o600` ?

sbchaos (approved):

Issue #222, opened 2022-03-15
lib.py to be copiled along with the job deployment as that might result into inconsistent behaviour otherwise.


sravankorumilli:
There is no need of bootstrap, just copying while deploying is sufficient.

sravankorumilli:
@sbchaos do we need to have bucket creation inside optimus, I believe we should expect bucket to be pre created

sbchaos:
We are adding `__lib.py` file to the bucket at the time of startup. We can check existence of this file and write at the time of deploy.

kushsharma:
I don't understand the reasoning? Did we just choose to remove the automation in favour of manual work?

kushsharma:
When does it goes out of sync? Also isn't we used to create bucket if it was not present or local folder if the directory was not present?

sravankorumilli:
Example and what we observed, there are changes in the lib.py which are not compatible with the dags, on new deployment the lib.py will get copied and all dags can't be loaded. It needs a redeployment of dags to be in sync with lib.py file.

sravankorumilli:
Currently too, optimus doesn't create the buckets, on local setup it does. But if we need to we can do that during the deployment. But it is better to not do the bucket creation as, that means giving extra privileges to optimus which isn't good.

PR #317, opened 2022-04-27
fix: deploy lib.py along with deploy job specs

sravankorumilli:
fixes issue #222


sbchaos (approved):

kushsharma:
I wanted to understand why are we uploading `lib` on each deployment?

PR #339, opened 2022-05-11
fix: bug when deleting secret

irainia:
This PR is to address a bug when we want to set a secret, where previously it's already deleted. Steps taken to encounter this error:

1. set secret with command `secret set [name] ...`
2. delete the secret with command `secret delete [name]`
3. re-run step 1. above and error should be encountered

The changes in this PR is to just add `Unscoped()` call function (reference)[https://gorm.io/docs/delete.html#Delete-permanently] during deletion to make Optimus delete it permanently.

### DIFF
--- a/store/postgres/secret_repository.go
+++ b/store/postgres/secret_repository.go
@@ -226,9 +226,9 @@ func (repo *secretRepository) Delete(ctx context.Context, project models.Project
 
 	var result *gorm.DB
 	if namespace.Name == "" {
-		result = query.Where("namespace_id is null").Delete(&Secret{})
+		result = query.Where("namespace_id is null").Unscoped().Delete(&Secret{})
 	} else {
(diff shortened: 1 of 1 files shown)


sravankorumilli (approved):

store/postgres/secret_repository.go
```diff
@@ -226,9 +226,9 @@ func (repo *secretRepository) Delete(ctx context.Context, project models.Project
 
 	var result *gorm.DB
 	if namespace.Name == "" {
-		result = query.Where("namespace_id is null").Delete(&Secret{})
+		result = query.Where("namespace_id is null").Unscoped().Delete(&Secret{})
```
sbchaos, line 229: Please add `Unscoped()` to the entire query.

```golang
query := repo.db.Unscoped().WithContext(ctx). <- Unscope at the query at line 223
```

irainia: updated

sbchaos (approved):

PR #350, opened 2022-05-19
feat: async job deployment

deryrahman:


models/job.go
```diff
@@ -329,7 +329,9 @@ type HTTPDependency struct {
 // JobService provides a high-level operations on DAGs
 type JobService interface {
 	// Create constructs a Job and commits it to a storage
-	Create(context.Context, NamespaceSpec, JobSpec) error
+	Create(context.Context, NamespaceSpec, JobSpec) (*JobSpec, error)
+	// BulkCreate save bulk of JobSpec within same namespace to persistent
+	BulkCreate(context.Context, NamespaceSpec, []JobSpec, progress.Observer) []JobSpec
```
irainia, line 334: Do we need to put it as the contract? As far as I see, it's only being used inside the service (apart from the unit test)

models/job.go
```diff
@@ -338,6 +340,8 @@ type JobService interface {
 	GetAll(context.Context, NamespaceSpec) ([]JobSpec, error)
 	// Delete deletes a job spec from all repos
 	Delete(context.Context, NamespaceSpec, JobSpec) error
+	// BulkDelete deletes bulk job specs
+	BulkDelete(context.Context, NamespaceSpec, []JobSpec, progress.Observer)
```
irainia, line 344: Similar to BulkCreate, this is only used inside the implementing service. Perhaps we don't need to export it.

models/job.go
```diff
@@ -329,7 +329,9 @@ type HTTPDependency struct {
 // JobService provides a high-level operations on DAGs
 type JobService interface {
 	// Create constructs a Job and commits it to a storage
-	Create(context.Context, NamespaceSpec, JobSpec) error
+	Create(context.Context, NamespaceSpec, JobSpec) (*JobSpec, error)
```
irainia, line 332: Why do we need to return it as pointer? When I check it in the implementation, we will get the value again. If possible, please make it consistent.

irainia (changes requested):

cmd/deploy/deploy.go
```diff
@@ -408,3 +416,199 @@ func (d *deployCommand) getResourceStreamClient(
 	}
 	return stream, nil
 }
+
+func (d *deployCommand) processJobDeploymentResponses(namespaceName string, stream pb.JobSpecificationService_DeployJobSpecificationClient) (string, error) {
```
sravankorumilli, line 420: Processing Job Deployment responses should be same in the refresh command too lets just reuses the same.

cmd/deploy/deploy.go
```diff
@@ -408,3 +416,199 @@ func (d *deployCommand) getResourceStreamClient(
...
+		if err != nil {
+			if errors.Is(err, io.EOF) {
+				break
+			}
+			return "", err
+		}
+
+		switch resp.Type {
+		case models.ProgressTypeJobDependencyResolution:
+			if !resp.GetSuccess() {
```
sravankorumilli, line 446: if we have both if else's and we are using ! then better swap the if else's

cmd/deploy/deploy.go
```diff
@@ -408,3 +416,199 @@ func (d *deployCommand) getResourceStreamClient(
...
+			if d.verbose {
+				// ordinary progress event
+				d.logger.Info(fmt.Sprintf("[%s] info '%s': %s", namespaceName, resp.GetJobName(), resp.GetValue()))
+			}
+		}
+	}
+	return "", nil
+}
+
+func (d *deployCommand) pollJobDeployment(ctx context.Context, jobSpecService pb.JobSpecificationServiceClient, deployID string) error {
```
sravankorumilli, line 571: this polling too will be common for all job deployments in all commands like job refresh lets reuse

job/service.go
```diff
@@ -817,6 +863,107 @@ func (srv *Service) resolveAndPersist(ctx context.Context, currentSpec models.Jo
...
+		if _, ok := requestedJobSpecs[jobName]; !ok {
+			deletedJobSpecs = append(deletedJobSpecs, existingJobSpec)
+		}
+	}
+
+	return deletedJobSpecs
+}
+
+func (*Service) jobSpecEqual(js1, js2 models.JobSpec) bool {
+	js2.ID = js1.ID
```
sravankorumilli, line 952: we can use reflect equal itself. instead of hash

models/progress.go
```diff
@@ -40,7 +43,24 @@ type (
 
 	// ProgressSavedJobDelete signifies that a raw
 	// job from a repository is being deleted
-	ProgressSavedJobDelete struct{ Name string }
+	ProgressSavedJobDelete struct {
```
sravankorumilli, line 46: why can't the events be just JobDeleteEvent, JobCreateEvent and JobModifyEvent

sravankorumilli (changes requested):

api/handler/v1beta1/job_spec.go
```diff
@@ -39,7 +38,13 @@ type JobSpecServiceServer struct {
 
 func (sv *JobSpecServiceServer) DeployJobSpecification(stream pb.JobSpecificationService_DeployJobSpecificationServer) error {
 	startTime := time.Now()
-	errNamespaces := []string{}
+
+	observers := sv.newObserverChain()
+	observers.Join(&jobDeploymentObserver{
```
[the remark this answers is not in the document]
arinda-arif: we map each observer with the response, if we prefer to have a single observer, that means we will need to unify the response (which is not preferred) or handle it in the observer part. also pre-processes in the refresh, job deploy, and job create are a little bit different (we are not finding diff and do save/delete operation). wdyt?

job/service.go
arinda-arif: we decided to use hash as it will solve the nil comparation

cmd/deploy/deploy.go
```diff
@@ -133,7 +140,11 @@ func (d *deployCommand) deploy(selectedNamespaces []*config.Namespace) error {
 	}
 
 	if !d.ignoreJobDeployment {
```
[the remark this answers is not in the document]
arinda-arif: sure, will do the check inside the deployResource and deployJobs.

api/handler/v1beta1/job_spec.go
```diff
@@ -50,78 +55,32 @@ func (sv *JobSpecServiceServer) DeployJobSpecification(stream pb.JobSpecificatio
...
+		deployID, err := sv.jobSvc.Deploy(stream.Context(), req.GetProjectName(), req.GetNamespaceName(), jobSpecs, observers)
 		if err != nil {
-			stream.Send(&pb.DeployJobSpecificationResponse{
-				Success: false,
-				Ack:     true,
-				Message: err.Error(),
-			})
-			errNamespaces = append(errNamespaces, req.NamespaceName)
+			observers.Notify(&models.ProgressJobDeploymentRequestCreated{Err: err})
+			sv.l.Warn(fmt.Sprintf("there's error while deploying namespaces: [%s]", req.NamespaceName))
```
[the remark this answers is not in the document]
arinda-arif: agreed to wrap the events to be sent to client-side (observer) with namespace details

sravankorumilli (approved):

Issue #342, opened 2022-05-17
update resources only when they are modified


sravankorumilli:
Resources can be updated after being created. Such an update process can happen through API calls or through the deployment process. Sometimes, no changes are done to them. But, during the process, they are re-deployed to the datastore, such as a bigquery. Every re-deployment to the datastore takes time.

Because of this, there needs to be a mechanism to avoid re-deployment to datastore if an update request is received but no changes are done. One example to be considered is by checking the difference between the incoming resource to the existing one. Only if they are different do we deploy to the datastore.

PR #351, opened 2022-05-20
feat: resource apply changes only if different

irainia:
This PR is to address #342 .

Changes:
* updated the implementation in service layer
* rework unit tests for the impacted methods

### DIFF
--- a/api/handler/v1beta1/observer.go
+++ b/api/handler/v1beta1/observer.go
@@ -65,20 +65,45 @@ func (obs *resourceObserver) Notify(e progress.Event) {
 	obs.mu.Lock()
 	defer obs.mu.Unlock()
 
-	evt, ok := e.(*datastore.EventResourceUpdated)
-	if ok {
-		resp := &pb.DeployResourceSpecificationResponse{
-			Success:      true,
-			Ack:          true,
-			ResourceName: evt.Spec.Name,
-		}
-		if evt.Err != nil {
-			resp.Success = false
-			resp.Message = evt.Err.Error()
-		}
+	var (
+		success               = true
+		resourceName, message string
+	)
 
+	var isEventRecognized bool
+	switch evt := e.(type) {
+	case *datastore.EventResourceCreated:
+		isEventRecognized = true
+		resourceName = evt.Spec.Name
+		if evt.Err != nil {
+			success = false
+			message = evt.Err.Error()
+		} else {
+			message = evt.String()
+		}
+	case *datastore.EventResourceUpdated:
+		isEventRecognized = true
+		resourceName = evt.Spec.Name
+		if evt.Err != nil {
+			success = false
+			message = evt.Err.Error()
+		} else {
+			message = evt.String()
+		}
+	case *datastore.EventResourceSkipped:
+		isEventRecognized = true
+		message = evt.String()
+	}
+
+	if isEventRecognized {
+		resp := &pb.DeployResourceSpecificationResponse{
+			Success:      success,
+			Ack:          true,
+			ResourceName: resourceName,
+			Message:      message,
+		}
 		if err := obs.stream.Send(resp); err != nil {
-			obs.log.Error("failed to send deploy spec ack", "spec name", evt.Spec.Name, "error", err)
+			obs.log.Error("failed to send deploy spec ack", "spec name", resourceName, "error", err)
 		}
 	}
 }
--- a/api/handler/v1beta1/observer.go
+++ b/api/handler/v1beta1/observer.go
@@ -61,6 +61,18 @@ type resourceObserver struct {
 	mu     *sync.Mutex
 }
 
+func NewResourceObserver(
+	stream pb.ResourceService_DeployResourceSpecificationServer,
+	log log.Logger,
+	mu *sync.Mutex,
+) progress.Observer {
+	return &resourceObserver{
+		stream: stream,
+		log:    log,
+		mu:     mu,
+	}
+}
+
 func (obs *resourceObserver) Notify(e progress.Event) {
 	obs.mu.Lock()
 	defer obs.mu.Unlock()
--- a/api/handler/v1beta1/observer.go
+++ b/api/handler/v1beta1/observer.go
@@ -70,10 +70,8 @@ func (obs *resourceObserver) Notify(e progress.Event) {
 		resourceName, message string
 	)
 
-	var isEventRecognized bool
 	switch evt := e.(type) {
 	case *datastore.EventResourceCreated:
-		isEventRecognized = true
 		resourceName = evt.Spec.Name
 		if evt.Err != nil {
 			success = false
@@ -82,7 +80,6 @@ func (obs *resourceObserver) Notify(e progress.Event) {
 			message = evt.String()
 		}
 	case *datastore.EventResourceUpdated:
-		isEventRecognized = true
 		resourceName = evt.Spec.Name
 		if evt.Err != nil {
 			success = false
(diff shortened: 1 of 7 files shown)


datastore/service.go
```diff
@@ -37,19 +39,30 @@ func (srv Service) GetAll(ctx context.Context, namespace models.NamespaceSpec, d
 func (srv Service) CreateResource(ctx context.Context, namespace models.NamespaceSpec, resourceSpecs []models.ResourceSpec, obs progress.Observer) error {
 	runner := parallel.NewRunner(parallel.WithLimit(ConcurrentLimit), parallel.WithTicket(ConcurrentTicketPerSec))
 	for _, resourceSpec := range resourceSpecs {
-		currentSpec := resourceSpec
-		repo := srv.resourceRepoFactory.New(namespace, currentSpec.Datastore)
+		incomingSpec := resourceSpec
+		repo := srv.resourceRepoFactory.New(namespace, incomingSpec.Datastore)
 		runner.Add(func() (interface{}, error) {
-			if err := repo.Save(ctx, currentSpec); err != nil {
+			proceed, err := srv.isProceedToSave(ctx, repo, incomingSpec)
```
sravankorumilli, line 45: cmiiw, I can observe that incoming spec will not have id and urn fields set so it will always save. Why not go with an equals comparision instead of hash instead of have has computation logic lets just go with equals comparision

datastore/service.go
```diff
@@ -37,19 +39,30 @@ func (srv Service) GetAll(ctx context.Context, namespace models.NamespaceSpec, d
...
+			if err != nil {
 				return nil, err
 			}
 
-			err := currentSpec.Datastore.CreateResource(ctx, models.CreateResourceRequest{
-				Resource: currentSpec,
+			if !proceed {
+				srv.notifyProgress(obs, &EventResourceCreated{
```
sravankorumilli, line 51: the event will be different if skipped update

datastore/service.go
```diff
@@ -68,19 +81,30 @@ func (srv Service) CreateResource(ctx context.Context, namespace models.Namespac
 func (srv Service) UpdateResource(ctx context.Context, namespace models.NamespaceSpec, resourceSpecs []models.ResourceSpec, obs progress.Observer) error {
 	runner := parallel.NewRunner(parallel.WithLimit(ConcurrentLimit), parallel.WithTicket(ConcurrentTicketPerSec))
 	for _, resourceSpec := range resourceSpecs {
-		currentSpec := resourceSpec
-		repo := srv.resourceRepoFactory.New(namespace, currentSpec.Datastore)
+		incomingSpec := resourceSpec
+		repo := srv.resourceRepoFactory.New(namespace, incomingSpec.Datastore)
 		runner.Add(func() (interface{}, error) {
```
sravankorumilli, line 86: same comments here too and there is code duplication any reason why we are not merging these flows.

sravankorumilli (changes requested):

datastore/service.go
irainia: updated by creating a new event

datastore/service.go
irainia: On this, it has been similar since the beginning (can be checked [here](https://github.com/odpf/optimus/blob/54f0784be99bd128a197ff0a1e4395c598d35efd/datastore/service.go)). If we it's required to be merged, then we have to also consider [this part](https://github.com/odpf/optimus/blob/54f0784be99bd128a197ff0a1e4395c598d35efd/ext/datastore/bigquery/bigquery.go) of the implementation since they are similar. If we can somehow merge the latter, then it shouldn't be much of issue about the former. If you are okay with it, further changes will be required to accommodate it.

irainia: this is updated as mentioned. regarding the issue mentioned around bigquery, it will be addressed [here](https://github.com/odpf/optimus/issues/397).

datastore/service.go
```diff
@@ -139,6 +111,61 @@ func (srv Service) DeleteResource(ctx context.Context, namespace models.Namespac
...
+	var errorSet error
+	for _, result := range runner.Run() {
+		if result.Err != nil {
+			errorSet = multierror.Append(errorSet, result.Err)
+		}
+	}
+	return errorSet
+}
+
+func (Service) isProceedToSave(ctx context.Context, repo store.ResourceSpecRepository, incomingSpec models.ResourceSpec) (bool, error) {
```
sbchaos, line 154: Can we rename the function from `isProceedToSave` to. `alreadyExists` and flip the boolean value from `proceed` to `exists`

sravankorumilli: Or why not have equals method on resource struct and ignore id and urn fields with the same approach. Just use the equals method   by inlining the function.

irainia: the intention is not to check whether the incoming exists or not. instead, it is to check whether the incoming is equal with the existing or not. I agree that using `isProceedToSave` should be changed. But, using `alreadyExists` also doesn't really reflect the intention. So, for now I just proposed to use `isEqualWithExisting`. Let me know if it works.

api/handler/v1beta1/observer.go
```diff
@@ -65,20 +65,45 @@ func (obs *resourceObserver) Notify(e progress.Event) {
...
 	defer obs.mu.Unlock()
 
-	evt, ok := e.(*datastore.EventResourceUpdated)
-	if ok {
+	var (
+		success               = true
+		resourceName, message string
+	)
+
+	var isEventRecognized bool
```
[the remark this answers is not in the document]
irainia: updated by removing it

datastore/service.go
```diff
@@ -35,65 +37,35 @@ func (srv Service) GetAll(ctx context.Context, namespace models.NamespaceSpec, d
...
-			})
-			return nil, err
-		})
-	}
-
-	var errorSet error
-	for _, result := range runner.Run() {
-		if result.Err != nil {
-			errorSet = multierror.Append(errorSet, result.Err)
+	execDatastore := func(rs models.ResourceSpec) error {
```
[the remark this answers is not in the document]
irainia: updated as mentioned

datastore/service.go
```diff
@@ -35,65 +37,35 @@ func (srv Service) GetAll(ctx context.Context, namespace models.NamespaceSpec, d
...
```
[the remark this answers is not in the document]
irainia: updated as mentioned

datastore/service.go
```diff
@@ -35,65 +37,35 @@ func (srv Service) GetAll(ctx context.Context, namespace models.NamespaceSpec, d
...
+		request := models.CreateResourceRequest{
+			Resource: rs,
+			Project:  namespace.ProjectSpec,
 		}
+		err := rs.Datastore.CreateResource(ctx, request)
```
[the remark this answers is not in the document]
irainia: this datastore is already there, even before this update (can be checked [here](https://github.com/odpf/optimus/blob/36a511102be17ad75d730068dffb8dc404a6c967/datastore/service.go#L47)). its purpose is to apply it to the provider, such as bigquery table

datastore/service.go
irainia: updated based on the feedback by @sravankorumilli

models/datastore_test.go
```diff
@@ -0,0 +1,134 @@
...
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/suite"
+
+	"github.com/odpf/optimus/models"
+)
+
+type ResourceSpecTestSuite struct {
```
sravankorumilli, line 11: we can avoid following the testify and use the suite functionality in go testing itself right?

models/datastore.go
```diff
@@ -24,18 +25,48 @@ func (r ResourceType) String() string {
 }
 
 type ResourceSpec struct {
-	ID        uuid.UUID
-	Version   int
-	Name      string
-	Type      ResourceType
+	// ID represents the unique identifier of a resource.
```
sravankorumilli, line 28: do we need comments here we can have the comments on the equals method that should be sufficient

datastore/service.go
```diff
@@ -139,6 +110,46 @@ func (srv Service) DeleteResource(ctx context.Context, namespace models.Namespac
...
+	resourceSpecs []models.ResourceSpec,
+	obs progress.Observer,
+	storeDatastore func(models.ResourceSpec) error,
+) error {
+	runner := parallel.NewRunner(parallel.WithLimit(ConcurrentLimit), parallel.WithTicket(ConcurrentTicketPerSec))
+	for _, incomingSpec := range resourceSpecs {
+		repo := srv.resourceRepoFactory.New(namespace, incomingSpec.Datastore)
+		runner.Add(func() (interface{}, error) {
+			existingSpec, err := repo.GetByName(ctx, incomingSpec.Name)
+			if err != nil && !errors.Is(err, store.ErrResourceNotFound) {
```
sravankorumilli, line 125: this will not handle the situation when error is other than resourcenotfound

models/datastore.go
irainia: updated as mentioned

PR #358, opened 2022-05-23
fix: add missing logs on job deployment and fix job refresh cli

arinda-arif:
1. Add logs of which project is currently being deployed and how long it has been started.
2. Add logs of which deployment is being canceled due to time out.
3. Remove unnecessary sprintf in job refresh CLI.
4. Treat job refresh stream EOF response as an error, as it should be unexpected. The expected EOF that Job Refresh expecting is when it receives the job deployment ID.

### DIFF
--- a/cmd/job/refresh.go
+++ b/cmd/job/refresh.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"io"
 	"time"
 
 	"github.com/odpf/salt/log"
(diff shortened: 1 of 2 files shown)


cmd/job/refresh.go
```diff
@@ -163,11 +162,7 @@ func (r *refreshCommand) getRefreshDeploymentID(stream pb.JobSpecificationServic
 	for {
 		response, err := stream.Recv()
 		if err != nil {
-			if errors.Is(err, io.EOF) {
-				break
-			}
-			streamError = err
-			return
+			return "", err
```
sravankorumilli, line 165: I remember we don't want to consider EOF as an error and hence implemented the if check right? has anything changed in our understanding?

sravankorumilli: If yes this is the implementation followed in many places.

PR #361, opened 2022-05-23
fix: initialize cli colors

arinda-arif:
Color initialization is being done in the root command's PersistentPreRun function. However, this function is not being run as it is overridden by the child's PersistentPreRun.

The color init is not necessarily placed in the root's PersistentPreRun, thus moving it to the root command's constructor.


cmd/commands.go
```diff
@@ -33,6 +33,9 @@ var disableColoredOut = false
 // unless the stdout/err is a tty, colors/progressbar should be disabled
 func New() *cli.Command {
 	disableColoredOut = !utils.IsTerminal(os.Stdout)
+	if !disableColoredOut {
```
sbchaos, line 36: Can we use the condition from line above, we are using negation of negation.

sbchaos (approved):

Issue #352, opened 2022-05-20
server crash when creating resource


irainia:
**Describe the bug**
This bug happens when we create a resource through the API call with an improper payload body.

**To Reproduce**

1. initiate the server
2. prepare payload, improper one (it can also be empty payload)
3. send a create resource request to the API, not through deployment client
4. server should crash when receiving this request

**Expected behavior**
The server should properly handle it. And instead of crashing, should return error.

PR #374, opened 2022-05-25
fix: server crash when accepting improper resource create request

irainia:
This PR is to address #352 . Even though the issue only mentioned about resource creation, but this solution should also address the same for updating resource. This is because the update process is similar to the creation.

### Changes

The followings are the changes done:

* update adapter to add further validation
* create the unit test to address such implementation
* update mock implementation to address the requirement

### Improvement

Based on this PR, there are some TODO that can be addressed through **separate cards**, as the changes might be bigger.

* add validation on the adapter initialization
* add more unit test on the adapter to address more cases
* update mock implementation to address two points above
* finding out the similar case where the server could crash based on the request from the user

### DIFF
--- a/api/handler/v1beta1/adapter.go
+++ b/api/handler/v1beta1/adapter.go
@@ -1,6 +1,7 @@
 package v1beta1
 
 import (
+	"errors"
 	"fmt"
 	"strings"
 	"time"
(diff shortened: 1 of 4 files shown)


api/handler/v1beta1/adapter.go
```diff
@@ -442,14 +443,24 @@ func ToResourceProto(spec models.ResourceSpec) (*pb.ResourceSpecification, error
 }
 
 func FromResourceProto(spec *pb.ResourceSpecification, storeName string, datastoreRepo models.DatastoreRepo) (models.ResourceSpec, error) {
+	if spec == nil {
```
sravankorumilli, line 446: @irainia cmiiw, because these flows are not handled, it is panicing and server is crashing is it? Just want to check ideally the panic should just kill the goroutine handling the request not the entire server, can we cross check on this, as this means this issue can happen even in the future for all unhandled scenarios.

irainia: @sravankorumilli , yes, it cashed because of panic. But, we didn't explicitly call any go routine to run it (such as [this](https://github.com/odpf/optimus/blob/e72edc7ef46cda9dd30691a5444f98f68e8bb1a7/api/handler/v1beta1/resource.go#L38)).

arinda-arif:
LGTM, although seeing a similar issue on job creation. We can consider fixing that in this or new PR should also be fine.

api/handler/v1beta1/adapter.go
sravankorumilli: @irainia we can use middleware as per [this](https://github.com/grpc-ecosystem/go-grpc-middleware/blob/master/recovery/examples_test.go) to recover from panics

irainia: updated and tested based on the suggestion

sravankorumilli (approved):

Issue #390, opened 2022-05-30
SuperExternalTaskSensor is poking incorrect upstream schedule in case the duration between two consecutive job instances is not equal


SumitAgrawal03071989:
**Describe the bug**
Currently, when the duration between two consecutive job instances is equal the SuperExternalTaskSensor works correctly, e.g. if the job is scheduled daily at 2 AM, the difference between two consecutive job instances is 24h, but in the case of jobs that are scheduled monthly(0 2 1 * *) or specific dates on month (0 2 2,11,19,26 * *) then the sensor does not poke correct upstream schedule.

**To Reproduce**
1. schedule any job with cron as monthly(0 2 1 * *) or custom dates on month(0 2 2,11,19,26 * *)
2. schedule daily / monthly job and set upstream dependency to a job created in step 1
3. Observe the SuperExternalSensor behavior by deploying the above set of jobs

**Expected behavior**
The sensor should select the upstream job to poke depending on the cron schedule correctly.

PR #391, opened 2022-05-30
fix: #390 SuperExternalTaskSensor fix upstream dependancy lookup 

SumitAgrawal03071989:


SumitAgrawal03071989:
Current logic: calculate duration (this duration is later used to calculate start and end date from scheduled time to execution time) as the difference between 1. next scheduled time and 2. next to next schedule time.

Proposed logic: As this^ duration may vary if the difference between two job instances is not always equal, we would need to calculate duration as the difference between the previously scheduled time and the currently scheduled time, adding pseudo-code for the proposed logic, 
note: this logic can simply be replaced if we can calculate previous schedule time from current time, but as this is not supported currently, adding this custom logic to calcualte previous schedule.

1. maintain an array (scheduleArr) to record the history of the cron schedule from the current scheduled time (passed as a parameter)
2. calculate the average duration between two job runs
    2.1 sum of duration between job runs
    2.2 averageDuration = sumOfDuration / numberOfIteration  
3. calculate a point in history by moving scheduled time to scheduled time - averageDuration * numberOfIteration time. 
4. record all schedules in scheduleArr by iterating over schedule cron.
5. the scheduleArr will contain the current schedule time and its previous time.
6. iterate scheduleArr to find the previous scheduled time.
7. calculate the duration as difference between current scheduled time and the previous scheduled time

sravankorumilli:
@SumitAgrawal03071989 can you provide tests for the same scenarios

sravankorumilli (changes requested):
Can you provide tests for the same scenarios

PR #403, opened 2022-06-08
fix: config read from filename

deryrahman:


config/loader.go
```diff
@@ -14,8 +14,8 @@ import (
 
 const (
 	ErrFailedToRead       = "unable to read optimus config file %v (%s)"
-	DefaultFilename       = "optimus"
-	DefaultConfigFilename = "config" // default file name for server config
+	DefaultFilename       = "optimus.yaml"
+	DefaultConfigFilename = "config.yaml" // default file name for server config
 	DefaultFileExtension  = "yaml"
```
sravankorumilli, line 19: what is the significance of the defaultfileextension then.

sravankorumilli, line 19: what is the significance of the defaultfileextension then.

deryrahman: Just checked that,`DefaultFileExtension` is used to define the type of parser.

For example, we have config yaml config file, but the name is `optimus.custom`. So we can use `yaml` parser to parse file `optimus.custom`

```go
DefaultFilename       = "optimus.custom"
DefaultFileExtension  = "yaml"
```

sravankorumilli (approved):

config/loader.go
sravankorumilli: hmmm the salt config says filename without extension and it was working without file extension as well. but only when both file name with and without extension present it is loading the file without extension.

deryrahman: yes, I think what it says is different with the source code [here](https://github.com/spf13/viper/blob/2b83d46130080293d70212603c78c512410eb2db/viper_go1_15.go#L39), hence there's a mismatch

PR #404, opened 2022-06-13
feat : executor boot process via init-container

SaikumarChintada:
Executor Boot Process:
* source env : 
Airlfow worker fetches assets from optimus server and adds env variables (job env and secrets) to the job POD.
* fetch assets : 
Init container fetches assets from optimus and writes to a shared volume for the executor container to pick it up.


ext/scheduler/airflow2/resources/base_dag.py
```diff
@@ -99,8 +99,51 @@
 )
 {{- end }}
 
+JOB_DIR = "/data"
+IMAGE_PULL_POLICY="IfNotPresent"
+INIT_CONTAINER_IMAGE="optimus:latest" # inject from optimus config ?
```
sravankorumilli, line 104: Init container image version should be based on the release version, such that it is compatible with the server deployed.

ext/scheduler/airflow2/resources/expected_compiled_template.py
```diff
@@ -136,24 +194,31 @@
...
-        k8s.V1EnvVar(name="NAMESPACE",value='bar-namespace'),
+    env_vars=executor_env_vars,
+    reattach_on_restart=True,
+    volume_mounts=asset_volume_mounts,
+    volumes=[volume],
+    init_containers=[init_container_predator],
+)
+
+
+init_container_hook__dash__for__dash__fail = k8s.V1Container(
```
sravankorumilli, line 205: why do we need this?

PR #406, opened 2022-06-14
feat: provide api related to optimus sensor

### LINKED ISSUE #399: Provide APIs to get JobSpec and JobRuns based on Resource URN
1. Provide API to fetch `JobSpec` based on resource URN. This will be used on Optimus Sensor creation, to identify whether the particular resource needs a cross Optimus sensor or can be ignored.
```(GET) {optimus_host}/api/v1beta1/jobs```
    _Params: resource_destination_
2. Provide API to fetch `JobRuns` based on resource URN. This will be used by the Optimus Sensor to check the run status of the particular resource given the date range.
    ```(GET) {optimus_host}/api/v1beta1/jobruns```
    _Params: resource_destination, dstart, dend_
    Response:
    ```json
     {
        "success" : true,		
        "job_runs" : [
            {
                "scheduled_at" : "2022-01-01 02:00:00",
                "state" : "failed"
            }
        ],
        "message" : ""
    }
    ```

### PR DESCRIPTION
**Pull request by SumitAgrawal03071989**
fix #399

### CODE REVIEW

File: store/postgres/inter_project_job_spec_repository.go [hunk 1]
```diff
@@ -0,0 +1,64 @@
...
+		query = query.Where("project_id = ?", project.ID)
+	}
+	if jobName != "" {
+		query = query.Where("name = ?", jobName)
+	}
+	if resourceDestination != "" {
+		query = query.Where("destination = ?", resourceDestination)
+	}
+	if err := query.Find(&jobs).Error; err != nil {
+		if errors.Is(err, gorm.ErrRecordNotFound) {
```
arinda-arif, line 39: We can remove this as in gorm v2, an empty result in the `Find` call will not return `ErrRecordNotFound`

store/postgres/inter_project_job_spec_repository.go
```diff
@@ -0,0 +1,64 @@
...
+	"context"
+	"errors"
+
+	"gorm.io/gorm"
+
+	"github.com/odpf/optimus/models"
+	"github.com/odpf/optimus/store"
+)
+
+type interProjectJobSpecRepository struct {
```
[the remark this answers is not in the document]
irainia: Agreed on this, we should redefine them later.

store/store.go
```diff
@@ -39,6 +39,11 @@ type ProjectJobSpecRepository interface {
 	GetJobNamespaces(ctx context.Context) (map[string][]string, error)
 }
 
+// InterProjectJobSpecRepository represents a storage interface for Job specification
+type InterProjectJobSpecRepository interface {
+	GetWithFilters(ctx context.Context, projectName, jobName, resourceDestination string) ([]models.JobSpec, error)
```
irainia, line 44: Instead of having one function to accepts all the parameters, we suggest to have it like this:
* we can split the function into two, like `GetByJobName` with return `[]models.JobSpec` and `GetByDestination` with return `models.JobSpec` as it is unique and expected to return only one (of course all of them with error as well in the return)
* we can put the responsibilities of selecting which repository to be called to service layer. such case can be:
  * get by project name and job name, we can use `ProjectJobSpecRepository.GetByName`
  * get by project name, we can use `ProjectJobSpecRepository.GetAll`
  * get by job name (only) is already mentioned in the first point
  * get by destination (only) is already mentioned in the first point
  * get by project name and destination, we don't need it as destination is unique
  * get by job name and destination, we don't need it as destination is unique
  * get by project name, job name, and destination, we don't need it as destination is unique
  * if none are set (meaning empty), then the service should return error (ensuring at least one being set)

cc: @arinda-arif

models/job.go
```diff
@@ -361,6 +361,8 @@ type JobService interface {
 	Deploy(context.Context, string, string, []JobSpec, progress.Observer) (DeploymentID, error)
 	// GetDeployment getting status and result of job deployment
 	GetDeployment(ctx context.Context, deployID DeploymentID) (JobDeployment, error)
+	// GetWithFilts gets the jobspec based on projectName, jobName, resourceDestination filters.
+	GetWithFilters(ctx context.Context, projectName string, jobName string, resourceDestination string) ([]JobSpec, error)
```
irainia, line 365: Please change it to:

```go
type JobService interface {
    ...
    GetByFilter(ctx context.Context, filter JobSpecFilter) ([]JobSpec, error)
    ...
}
```

and for the `JobSpecFilter`:

```go
type JobSpecFilter struct {
    ProjectName               string
    JobName                     string
    ResourceDestination  string
}
```

With this, if there's any change to the filter, we don't need to update the interface. cc: @arinda-arif

irainia (changes requested):

store/store.go
SumitAgrawal03071989: So request with project name and destination will be served as empty jobsspecs as its not needed?

SumitAgrawal03071989:
@arinda-arif @irainia I updated the review changes, could you please review

job/service.go
```diff
@@ -170,6 +171,43 @@ func (srv *Service) GetByName(ctx context.Context, name string, namespace models
 	return jobSpec, nil
 }
 
+func (srv *Service) GetByFilter(ctx context.Context, filter models.JobSpecFilter) ([]models.JobSpec, error) {
+	if filter.ResourceDestination != "" {
+		jobSpec, err := srv.interProjectJobSpecRepository.GetJobByResourceDestination(ctx, filter.ResourceDestination)
+		if err != nil {
+			return []models.JobSpec{}, err
```
irainia, line 178: Can you change it to return `nil, err` instead? I believe the slice won't be needed if there's an error. This is to reduce unnecessary memory allocation.

store/postgres/inter_project_job_spec_repository.go
```diff
@@ -0,0 +1,49 @@
...
+	db      *gorm.DB
+	adapter *JobSpecAdapter
+}
+
+func (repo interProjectJobSpecRepository) GetJobByName(ctx context.Context, jobName string) ([]models.JobSpec, error) {
+	var jobs []Job
+	var specs []models.JobSpec
+	if err := repo.db.WithContext(ctx).Preload("Namespace").Preload("Project").Where("name = ?", jobName).Find(&jobs).Error; err != nil {
+		return []models.JobSpec{}, err
```
irainia, line 20: we can apply here as well about the nil

irainia (changes requested):

job/service.go
irainia: cc: @arinda-arif

store/postgres/inter_project_job_spec_repository.go
```diff
@@ -0,0 +1,49 @@
...
+			return specs, err
+		}
+		specs = append(specs, adapt)
+	}
+	return specs, nil
+}
+
+func (repo interProjectJobSpecRepository) GetJobByResourceDestination(ctx context.Context, resourceDestination string) (models.JobSpec, error) {
+	var job Job
+	if err := repo.db.WithContext(ctx).Preload("Namespace").Preload("Project").Where("destination = ?", resourceDestination).First(&job).Error; err != nil {
```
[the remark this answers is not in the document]
irainia: cc: @arinda-arif

cmd/server/optimus.go
```diff
@@ -216,9 +216,11 @@ func (s *OptimusServer) setupHandlers() error {
 	projectSecretRepo := postgres.NewSecretRepository(s.dbConn, s.appKey)
 
 	dbAdapter := postgres.NewAdapter(models.PluginRegistry)
+	jobSpecAdapter := postgres.NewAdapter(models.PluginRegistry)
```
sbchaos, line 219: Please use the adapter `dbAdapter` defined at line 218

store/postgres/inter_project_job_spec_repository.go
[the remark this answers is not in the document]
SumitAgrawal03071989: I think project will be needed to get the job run. removing namespace from preload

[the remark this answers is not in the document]
SumitAgrawal03071989: Actually it will break while converting `postgres.job` to `models.jobspec` if we exclude either project or namespace I checked other queries and it follows the same.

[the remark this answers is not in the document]
irainia: I see, if that's the case then no need to change it. We can resolve this one.

store/postgres/inter_project_job_spec_repository.go
```diff
@@ -0,0 +1,49 @@
+package postgres
```
irainia, line 1: Can you also add some unit tests for this file? The current one doesn't pass the pipeline.

irainia (changes requested):

irainia (approved):
LGTM

arinda-arif (approved):
LGTM. @SumitAgrawal03071989 You can update the `PROTON_COMMIT` too as the PR in Proton has been merged.

If nothing else from @sbchaos, we can merge this.

sbchaos (approved):

PR #433, opened 2022-07-05
feat: window Playground command

Tharun1718333:
Added a playground command for getting the start and end time of the window from the user input


sravankorumilli:
@Tharun1718333 there is a need to reuse the implementation and integrate with the existing code. Can you work on that and  there is a need for improving the modelling too.

PR #434, opened 2022-07-05
feat : add plugin manager

smarch-int:
New Features: 
* `optimus serve --install-plugins`
* `optimus plugin install` [hidden]
* sample plugin config within optimus server config
```yaml
plugin:
  dir: "." # plugin discovery logic needs to be as is (support current client plugins)
  artifacts:
    # refer : https://github.com/hashicorp/go-getter
    - ../transformers/dist/transformers_0.1.0_macos_arm64.tar.gz
    - ../transformers/dist/transformers_0.1.0_windows_x86_64.zip
    - https://github.com/odpf/optimus/releases/download/v0.2.5/optimus_0.2.5_linux_arm64.tar.gz
```


go.mod
```diff
@@ -55,7 +55,7 @@ require (
 	golang.org/x/tools v0.1.10 // indirect
 	google.golang.org/api v0.56.0
 	google.golang.org/genproto v0.0.0-20210903162649-d08c68adba83
-	google.golang.org/grpc v1.46.0
```
sravankorumilli, line 58: why this change

go.mod
```diff
@@ -43,7 +43,7 @@ require (
 	github.com/spf13/viper v1.8.1
 	github.com/stretchr/testify v1.7.1
 	github.com/xlab/treeprint v1.1.0
-	go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc v0.32.0
+	go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc v0.25.0
```
sravankorumilli, line 46: why this downgrage

cmd/plugin/plugin_manager.go
```diff
@@ -0,0 +1,62 @@
...
+		client: client,
+	}
+}
+
+type PluginManager struct {
+	logger log.Logger
+	client *getter.Client
+}
+
+func (p *PluginManager) Install(src, dst string) error {
```
sravankorumilli, line 40: just use variable arguments method instead of two methods

cmd/plugin/install.go
```diff
@@ -0,0 +1,46 @@
...
+	"github.com/odpf/optimus/config"
+)
+
+type installCommand struct {
+	logger       log.Logger
+	serverConfig *config.ServerConfig
+}
+
+// NewInstallCommand initializes plugin install command
+func NewInstallCommand(serverConfig *config.ServerConfig) *cobra.Command {
```
sravankorumilli, line 17: as discussed we need to remove this command

sravankorumilli (changes requested):

cmd/plugin/plugin.go
```diff
@@ -5,11 +5,45 @@ import (
...
 	hPlugin "github.com/hashicorp/go-plugin"
+	"github.com/spf13/cobra"
 
 	"github.com/odpf/optimus/config"
 	oPlugin "github.com/odpf/optimus/plugin"
 )
 
+type pluginCommand struct {
+	configFilePath string
+	serverConfig   *config.ServerConfig
```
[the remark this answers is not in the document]
smarch-int: I get your point, but when I add another command in plugin then I have to move it back to plugin level again.
I think its better to have all deps injected from parent command.

sravankorumilli (approved):

sravankorumilli (approved):

PR #442, opened 2022-07-18
feat: window Playground command 

Tharun1718333:
Added a playground command for getting the start and end time of the window from the user input


cmd/playground/interactive_window.go
```diff
@@ -0,0 +1,213 @@
...
+
+	"github.com/charmbracelet/bubbles/textinput"
+	tea "github.com/charmbracelet/bubbletea"
+	"github.com/odpf/optimus/job"
+)
+
+// model contains a cursor to keep track of the pointer, two input fields for taking Size and Offset as input
+// model also contains a state which maintains both versions of windows and the sechdule time
+type model struct {
+	cursor      string
```
irainia, line 16: using `string` for `cursor` field is a bit confusing. we have to guess what is inside this `cursor`. this is also related when we use it under `genarateFields`. we don't know what is this `position` gonna be? you can create a different type for this, like `cursorType` or something.

cmd/playground/interactive_window.go
```diff
@@ -0,0 +1,213 @@
...
+		}
+	}
+	// update the values of the window versions in state for every modification from input
+	m.updateWindowVersions()
+	return m, nil
+}
+
+// we genarate a cursor if the value of the cursor matches with the current field name, else we pass a blank instead
+func (m model) genarateFields(position string, value string) string {
+	if m.cursor == position {
```
irainia, line 149: Looking at this implementation, I'm still not sure how this logic works.

cmd/playground/interactive_window.go
```diff
@@ -0,0 +1,213 @@
...
+			m.Offsetinput, _ = m.Offsetinput.Update(msg)
+		}
+	}
+	// update the values of the window versions in state for every modification from input
+	m.updateWindowVersions()
+	return m, nil
+}
+
+// we genarate a cursor if the value of the cursor matches with the current field name, else we pass a blank instead
+func (m model) genarateFields(position string, value string) string {
```
irainia, line 148: try to change this method to better reflect what it does. currently I'm not sure what this method does. if what you mean is to generate input fields view, then perhaps you can name it like `generateInputFieldsView` or other better name. With this name, I first assumed we will get a list of fields.

cmd/playground/interactive_window.go
```diff
@@ -0,0 +1,213 @@
...
+			m.Sizeinput, _ = m.Sizeinput.Update(msg)
+			m.Offsetinput, _ = m.Offsetinput.Update(msg)
+		}
+	}
+	// update the values of the window versions in state for every modification from input
+	m.updateWindowVersions()
+	return m, nil
+}
+
+// we genarate a cursor if the value of the cursor matches with the current field name, else we pass a blank instead
```
irainia, line 147: try to update or remove this comment, as this doesn't help me understand the method. like, which part is `field name` in the method.

cmd/playground/interactive_window.go
```diff
@@ -0,0 +1,213 @@
...
+	s.WriteString(m.Offsetinput.View())
+	s.WriteString("\n")
+	s.WriteString("TruncateTo       ")
+	s.WriteString(m.genarateFields("TruncateTo", m.state.windowv2.TruncateTo))
+	s.WriteString("\n")
+	s.WriteString("sechduled date   ")
+	s.WriteString(m.genarateSechduledDateView())
+	s.WriteString("\n\n\n             ") // extra space added for design reasons
+	// calculate the value of dstart and dend from the imported window versions
+	dstartv1, dendv1, err := m.state.windowv1.GetTimeRange(m.state.sechduledTime)
```
irainia, line 197: use `dStartV1` instead of `dstartv1` (this also for the `end` and the version 2 as well)

cmd/playground/playground.go
```diff
@@ -0,0 +1,25 @@
+package playground
+
+import (
+	"github.com/odpf/optimus/config"
```
irainia, line 4: arrange this import so internal optimus dep can be at bottom most

cmd/playground/window.go
```diff
@@ -0,0 +1,72 @@
+package playground
+
+import (
+	"time"
+
+	"github.com/odpf/optimus/job"
+)
+
+type state struct {
```
irainia, line 9: after checking, some of the responsibility of this `state` is handled directly by the `model` struct. if we want to retain this in a separate struct, then let this struct has its own responsiblity. like changing the value of `window1/2` or something instead of allowing `model` to do that. however, if `model` is still to handle this, then better merge this `state` struct to that.

cmd/playground/playground_window.go
```diff
@@ -0,0 +1,38 @@
+package playground
```
irainia, line 1: change the file name from `playground_window.go` to be just `window.go` since we already know that this is inside a `playground`

irainia (changes requested):
Some of the feedback is already mentioned in a separate conversation previously. We can have yet another separate session to resolve it if you think one or more of those feedbacks doesn't make sense or there are other better ways.

cmd/playground/window.go
```diff
@@ -0,0 +1,35 @@
...
+	cmd := &cobra.Command{
+		Use:   "window",
+		Short: "get dStart,dEnd by giving the window params",
+		RunE:  window.RunE,
+	}
+	return cmd
+}
+
+func (j *windowCommand) RunE(_ *cobra.Command, _ []string) error {
+	// starts the interactive session
```
irainia, line 26: Please remove this comment

cmd/playground/window_model.go
```diff
@@ -0,0 +1,207 @@
...
+	offsetInput.Placeholder = defaultSize
+	return &model{
+		cursor:      pointToSize,
+		state:       state{windowV1: windowV1, windowV2: windowV2, scheduledTime: time.Now()},
+		sizeInput:   sizeInput,
+		offsetInput: offsetInput,
+	}
+}
+
+// since we don not intend to keep any input after the process is killed we return nil
```
irainia, line 41: Please remove this comment

cmd/playground/window_model.go
```diff
@@ -0,0 +1,207 @@
...
+func (*model) Init() tea.Cmd {
+	return nil
+}
+
+// this handles motion of the pointer when we hit arrow-up key
+func (m *model) handleUp() {
+	switch m.cursor {
+	case pointToOffset:
+		// here we are shifting from Offsetinput to Sizeinput so we have to stop updating the Offsetinput(hence Offset.Blur())
```
irainia, line 50: please update the letter case, for example changing `Offsetinput` to `offsetInput` or `offset input` or anything else, as long as it is consistent with the message to the user or with the variables in the implementation

cmd/playground/window_model.go
```diff
@@ -0,0 +1,207 @@
...
+		offsetInput: offsetInput,
+	}
+}
+
+// since we don not intend to keep any input after the process is killed we return nil
+func (*model) Init() tea.Cmd {
+	return nil
+}
+
+// this handles motion of the pointer when we hit arrow-up key
```
irainia, line 46: avoid using `this`, use the name of the method or function, such as in this case `handleUp handles ...`

cmd/playground/window_model.go
```diff
@@ -0,0 +1,207 @@
...
+		m.offsetInput.Blur()
+		m.cursor = pointToTruncate
+	case pointToSize:
+		m.offsetInput.Focus()
+		m.sizeInput.Blur()
+		m.cursor = pointToOffset
+	}
+}
+
+// handles left and right arrow key movements
```
irainia, line 77: similar comment as the `handleUp`

cmd/playground/window_model.go
```diff
@@ -0,0 +1,207 @@
...
+	case PointToDay:
+		m.cursor = pointToMonth
+	case pointToMinute:
+		m.cursor = PointToDay
+	case pointToHour:
+		m.cursor = pointToMinute
+	}
+}
+
+func (m *model) handleLeft() {
```
irainia, line 91: if the other `handle...` has comment, better put it as well here. if not, then we can remove all comments related to `handle...`. this is to be more consistent

cmd/playground/window_model.go
```diff
@@ -0,0 +1,207 @@
...
+		m.cursor = pointToHour
+	case PointToDay:
+		m.cursor = pointToMinute
+	case pointToMonth:
+		m.cursor = PointToDay
+	case pointToYear:
+		m.cursor = pointToMonth
+	}
+}
+func (m *model) handleIncrease() {
```
irainia, line 103: similar with the previous comment

cmd/playground/window_model.go
```diff
@@ -0,0 +1,207 @@
...
+	switch m.cursor {
+	case pointToTruncate:
+		m.state.IncrementTruncate()
+	case pointToYear, pointToMonth, PointToDay, pointToHour, pointToMinute:
+		m.state.IncrementTime(m.cursor.getStringValue())
+	}
+}
+func (m *model) handleDecrease() {
```
irainia, line 111: similar with the previous comment

cmd/playground/window_model.go
```diff
@@ -0,0 +1,207 @@
...
+	switch m.cursor {
+	case pointToTruncate:
+		m.state.DecrementTruncate()
+	case pointToYear, pointToMonth, PointToDay, pointToHour, pointToMinute:
+		m.state.DecrementTime(m.cursor.getStringValue())
+	}
+}
+
+func (m *model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
+	// switch currMsg := msg.(type) {
```
irainia, line 121: Please remove this comment

cmd/playground/window_model.go
```diff
@@ -0,0 +1,207 @@
...
+	// case tea.KeyMsg:
+	currMsg := reflect.TypeOf(msg)
+	if currMsg.String() == "tea.KeyMsg" {
```
irainia, line 124: instead of this, use the following:
```go
if currMsg.String() != "tea.KeyMsg" {
    return m, nil
}

switch fmt.Sprintf("%s", msg) {
...
```

This is to express that we care only if the message is "tea.KeyMsg", which in this case will be put as the top level flow, rather than lower level (or nested level).

cmd/playground/window_model.go
```diff
@@ -0,0 +1,207 @@
...
+		case "left":
+			m.handleLeft()
+		case "right":
+			m.handleRight()
+		case "shift+up", "shift+right":
+			m.handleIncrease()
+		case "shift+down", "shift+left":
+			m.handleDecrease()
+		case "M", "h", "-", "1", "2", "3", "4", "5", "6", "7", "8", "9", "0", "backspace":
+			// to get updated the input fields must be focussed since we handle the Focus and Blur we can directly update them here
```
irainia, line 141: use multiple lines if it's too long. avoid making the code too long to the right unless you have strong reasong of doing so.

cmd/playground/window_model.go
```diff
@@ -0,0 +1,207 @@
...
+		var s strings.Builder
+		s.WriteString("[")
+		s.WriteString(value)
+		s.WriteString("]")
+		return s.String()
+	}
+	return " " + value
+}
+
+// we genare a string representing the sechduled time , which also adds a cursor if it is pointing to the any of the fields in sechdueld date
```
irainia, line 161: avoid any subject like I, we, you, they, etc. Instead, use the name of the method like `genarateSechduledDateView generates a string ...`

cmd/playground/window_model.go
```diff
@@ -0,0 +1,207 @@
...
+	s.WriteString(m.genarateCursorIfPointed(pointToHour.getStringValue(), strconv.Itoa(m.state.scheduledTime.Hour())))
+	s.WriteString(":")
+	s.WriteString(m.genarateCursorIfPointed(pointToMinute.getStringValue(), strconv.Itoa(m.state.scheduledTime.Minute())))
+	s.WriteString(m.genarateCursorIfPointed(PointToDay.getStringValue(), strconv.Itoa(m.state.scheduledTime.Day())))
+	s.WriteString(m.genarateCursorIfPointed(pointToMonth.getStringValue(), m.state.scheduledTime.Month().String()))
+	s.WriteString(m.genarateCursorIfPointed(pointToYear.getStringValue(), strconv.Itoa(m.state.scheduledTime.Year())))
+	return s.String()
+}
+
+// this will update the values of Size and offset of the window versions present in state to the new values taken from the input
```
irainia, line 173: the same comment. just make sure to tidy it up.

cmd/playground/window_model.go
irainia: Also, the method name is typo, should be `generate...` not `genarate...`

cmd/playground/window_model.go
```diff
@@ -0,0 +1,207 @@
...
+
+// this will  be invoked for every update
```
irainia, line 175: no need for comment

cmd/playground/window_model.go
```diff
@@ -0,0 +1,207 @@
...
+	s.WriteString("\n")
+	s.WriteString("Offset\t\t")
+	s.WriteString(m.offsetInput.View())
+	s.WriteString("\n")
+	s.WriteString("TruncateTo\t ")
+	s.WriteString(m.genarateCursorIfPointed(pointToTruncate.getStringValue(), m.state.windowV2.TruncateTo))
+	s.WriteString("\n")
+	s.WriteString("sechduled date   ")
+	s.WriteString(m.genarateSechduledDateView())
+	s.WriteString("\n\n\n             ")
```
irainia, line 189: try to use `\t` for any whitespace to make it easier to decide the number of whitespaces to be used.

cmd/playground/window_model.go
```diff
@@ -0,0 +1,207 @@
...
+	// calculate the value of dstart and dend from the imported window versions
+	dStartV1, dEndV1, err := m.state.windowV1.GetTimeRange(m.state.scheduledTime)
+	if err != nil {
+		// if the window v1 cannot get time range from the given input it displays this error
+		s.WriteString("current window version does not support the above input")
+	} else {
+		s.WriteString("dStartV1 :" + dStartV1.Format("15:04 2006-01-02") + "     dEndV1 :" + dEndV1.Format("15:04 2006-01-02"))
+	}
+	s.WriteString("\n             ") // extra space added for design reasons
```
irainia, line 198: no need for comment here, also, consider using `\t`

cmd/playground/window_state.go
```diff
@@ -0,0 +1,75 @@
...
+	"github.com/odpf/optimus/job"
+)
+
+type state struct {
+	windowV1      job.WindowV1
+	windowV2      job.WindowV2
+	scheduledTime time.Time
+}
+
+// increment the component of the time which the cursor is pointed at
```
irainia, line 15: same comment for this comment

cmd/playground/window_state.go
```diff
@@ -0,0 +1,75 @@
...
+	case PointToDay.getStringValue():
+		s.scheduledTime = s.scheduledTime.AddDate(0, 0, 1)
+	case pointToMonth.getStringValue():
+		s.scheduledTime = s.scheduledTime.AddDate(0, 1, 0)
+	case pointToYear.getStringValue():
+		s.scheduledTime = s.scheduledTime.AddDate(1, 0, 0)
+	}
+}
+
+// decrement the component of the time which the cursor is pointed at
```
irainia, line 31: same comment in term of comment and implementation as in `IncrementTime` method

cmd/playground/window_state.go
```diff
@@ -0,0 +1,75 @@
...
+)
+
+type state struct {
+	windowV1      job.WindowV1
+	windowV2      job.WindowV2
+	scheduledTime time.Time
+}
+
+// increment the component of the time which the cursor is pointed at
+func (s *state) IncrementTime(increaseBy string) {
```
irainia, line 16: Please change the method name, for example, to `IncrementScheduleTimeOn` and the variable name `cursor cursorPointer`. This will make it easier to read that, we increase schedule time based on the cursor position. And if we do that, we will no longer need `getStringValue` method

cmd/playground/window_state.go
```diff
@@ -0,0 +1,75 @@
...
+	case PointToDay.getStringValue():
+		s.scheduledTime = s.scheduledTime.AddDate(0, 0, -1)
+	case pointToMonth.getStringValue():
+		s.scheduledTime = s.scheduledTime.AddDate(0, -1, 0)
+	case pointToYear.getStringValue():
+		s.scheduledTime = s.scheduledTime.AddDate(-1, 0, 0)
+	}
+}
+
+// change the value of truncate
```
irainia, line 47: same method for name

cmd/playground/window_state.go
```diff
@@ -0,0 +1,75 @@
...
+func (s *state) IncrementTruncate() {
```
irainia, line 48: If no one uses this outside, just make it as unexported (small `I`) and please update the others as well

cmd/playground/window_state.go
```diff
@@ -0,0 +1,75 @@
...
+	case "w":
+		s.windowV2.TruncateTo = "M"
+	case "d":
+		s.windowV2.TruncateTo = "w"
+	case "h":
+		s.windowV2.TruncateTo = "d"
+	}
+}
+
+func (s *state) DecrementTruncate() {
```
irainia, line 59: same comment

go.mod
```diff
@@ -65,7 +65,16 @@ require (
 	gorm.io/gorm v1.21.16
 )
 
-require go.uber.org/multierr v1.7.0
+require (
+	github.com/charmbracelet/bubbles v0.13.0
+	github.com/charmbracelet/bubbletea v0.22.0
+)
```
irainia, line 71: try to merge these two

irainia (changes requested):
These are minors feedback, let me know if discussion is required

cmd/playground/window_model.go
irainia: this is still not updated, why already resolved?

cmd/playground/window_state.go
```diff
@@ -0,0 +1,72 @@
...
+	"github.com/odpf/optimus/job"
+)
+
+type state struct {
+	windowV1      job.WindowV1
+	windowV2      job.WindowV2
+	scheduledTime time.Time
+}
+
+func (s *state) incrementScheduleTimeOn(increaseBy cursorPointer) {
```
irainia, line 15: change the name from `increaseBy` into `cursor`

cmd/playground/window_state.go
```diff
@@ -0,0 +1,72 @@
...
+	case PointToDay:
+		s.scheduledTime = s.scheduledTime.AddDate(0, 0, 1)
+	case pointToMonth:
+		s.scheduledTime = s.scheduledTime.AddDate(0, 1, 0)
+	case pointToYear:
+		s.scheduledTime = s.scheduledTime.AddDate(1, 0, 0)
+	}
+}
+
+func (s *state) decrementScheduleTimeOn(decreaseBy cursorPointer) {
```
irainia, line 30: also this, change `decreaseBy` into `cursor`

cmd/playground/window_state.go
```diff
@@ -0,0 +1,72 @@
...
+	switch s.windowV2.TruncateTo {
+	case "d":
+		s.windowV2.TruncateTo = "h"
+	case "w":
+		s.windowV2.TruncateTo = "d"
+	case "M":
+		s.windowV2.TruncateTo = "w"
+	}
+}
+func (s *state) updateWindowparameters(size string, offset string) {
```
irainia, line 66: if we want to use `parameters`, at least use capital `P` for `Parameters`

cmd/playground/window_state.go
```diff
@@ -0,0 +1,72 @@
...
+	case PointToDay:
+		s.scheduledTime = s.scheduledTime.AddDate(0, 0, -1)
+	case pointToMonth:
+		s.scheduledTime = s.scheduledTime.AddDate(0, -1, 0)
+	case pointToYear:
+		s.scheduledTime = s.scheduledTime.AddDate(-1, 0, 0)
+	}
+}
+
+func (s *state) incrementTruncate() {
```
irainia, line 45: use `TruncateTo` instead of `Truncate`, to make it more consistent

cmd/playground/window_state.go
```diff
@@ -0,0 +1,72 @@
...
+	case "w":
+		s.windowV2.TruncateTo = "M"
+	case "d":
+		s.windowV2.TruncateTo = "w"
+	case "h":
+		s.windowV2.TruncateTo = "d"
+	}
+}
+
+func (s *state) decrementTruncate() {
```
irainia, line 56: similar with above, use `TruncateTo`

irainia (changes requested):
Please update this as well, let me know if discussion is required.

cmd/playground/window_state.go
irainia: this is actually already mentioned in the previous comment. let me know if it's better to be implemented or there's better idea.

cmd/playground/window_state.go
irainia: this is actually already mentioned in the previous comment. let me know if it's better to be implemented or there's better idea.

irainia:
Also, the current pipeline failed. Try to resolve the `lint` jobs and ignore the `unite-test` one for now.

irainia:
Also, the current pipeline failed. Try to resolve the `lint` jobs and ignore the `unite-test` one for now.

sravankorumilli:
@Tharun1718333 & @irainia can you resolve the conflicts.

irainia:
@Tharun1718333 we can have a separate discussion to resolve this, just let me know

irainia:
Closing this PR as it is addressed by [this PR](https://github.com/odpf/optimus/pull/550) with some updates to also resolve the conflicts.

PR #446, opened 2022-07-19
refactor: simplify observer response for job deployment

deryrahman:


core/sender/log_status_deploy_job.go
```diff
@@ -0,0 +1,20 @@
+package sender
+
+import (
+	pb "github.com/odpf/optimus/api/proto/odpf/optimus/core/v1beta1"
+)
+
+type deployJobLogStatus struct {
```
sravankorumilli, line 7: we can have multiple Writers/Senders  DeployJobSpecificationResponseWriter/Sender
which can implement the interface for Sending Logs & we can have a LogBuilder which builds the proto LogStatus.

sravankorumilli: Lets add the responsibility of writing deploymentID as part of this method itself.

deryrahman: sure, one thing that I concern when I made a changes, should we remove the helper or not? I remember passing the toggle / flag in the function is considered as a bad design https://martinfowler.com/articles/feature-toggles.html

which one is better?

```go
// using helper
sender.SendSuccessMessage(deploySender, "success")
sender.SendErrorMessage(deploySender, "error")

// or
deploySender.SendLogStatus(LevelInfo, "success")
deploySender.SendLogStatus(LevelError, "error")
```

I feel method typing "SendSuccess" is more convenient as it reduces mistyping than passing toggle / flag "LevelInfo" in func parameter. But it's opinionated

api/handler/v1beta1/job_spec.go
```diff
@@ -38,39 +39,37 @@ type JobSpecServiceServer struct {
...
 			continue
 		}
 
-		sv.l.Info(fmt.Sprintf("deployID %s holds deployment for namespace %s\n", deployID.UUID().String(), req.NamespaceName))
-		observers.Notify(&models.ProgressJobDeploymentRequestCreated{DeployID: deployID})
+		successMsg := fmt.Sprintf("deployID %s holds deployment for namespace %s\n", deployID.UUID().String(), req.NamespaceName)
+		sv.l.Info(successMsg)
+		responseWriter.Write(writer.LogLevelInfo, successMsg)
+
+		resp := pb.DeployJobSpecificationResponse{DeploymentId: deployID.UUID().String()}
```
sravankorumilli, line 71: you can use the sendDeploymentId here

sravankorumilli (changes requested):

PR #452, opened 2022-07-20
feat : Yaml based Plugin for CLI features

smarch-int:
Below are powered via YAML instead of binaries:
1. Survey questions for job_create & job_adhook
2. Plugin Info
3. https://github.com/odpf/optimus/pull/530

Changes include:
- model `Plugin` to contain `YamlMod` reference to `PluginSpec` which implements `CommandLineMod` Interface.
```go
// representation of a yaml plugin
type AutoGenerated struct {
	Name          string `yaml:"name"`
	Description   string `yaml:"description"`
	Plugintype    string `yaml:"plugintype"`
	Pluginversion string `yaml:"pluginversion"`
	Image         string `yaml:"image"`
	Secretpath    string `yaml:"secretpath"`
	Questions     []struct {
		Name            string `yaml:"name"`
		Prompt          string `yaml:"prompt"`
		Help            string `yaml:"help"`
		Regexp          string `yaml:"regexp"`
		Validationerror string `yaml:"validationerror"`
		Minlength       int    `yaml:"minlength"`
		Maxlength       int    `yaml:"maxlength,omitempty"`
	} `yaml:"questions"`
	Defaultassets []struct {
		Name  string `yaml:"name"`
		Value string `yaml:"value"`
	} `yaml:"defaultassets"`
	Defaultconfig []struct {
		Name  string `yaml:"name"`
		Value string `yaml:"value"`
	} `yaml:"defaultconfig"`
}

// to enforce yaml plugin implementation
type YamlMod interface {
	BasePlugin
	CommandLineMod
}

// discuss issue 
// plugin is not an interface
// plugin is a struct that has attributes that refers to impl. of interfaces
type Plugin struct {
	Base          BasePlugin
	CLIMod        CommandLineMod
	DependencyMod DependencyResolverMod
	YamlMod       YamlMod
}


type PluginRepository interface {
	AddYaml(YamlMod) error                                            // yaml plugin 
	Add(BasePlugin, CommandLineMod, DependencyResolverMod) error       // binary plugin 
	
        // more..
}
```
- Plugin discovery:
  -  load binary plugins first
  -  then load yaml plugin only if a binary (with the same name) is not loaded
- `Plugin.YamlMod` to drive survey (job create & add hook);   instead of CLIMod if plugin is loaded as yaml
- Sorting Plugins listing
- Add yaml plugin validation (strict)
`optimus plugin validate --path {{file ||  dir}} `


cmd/plugin/validate.go
```diff
@@ -0,0 +1,108 @@
...
+	return cmd
+}
+
+func (v *validateCommand) PreRunE(_ *cobra.Command, _ []string) error {
+	v.logger = logger.NewClientLogger(v.serverConfig.Log)
+	return nil
+}
+
+func (v *validateCommand) validateFile(pluginPath string) error {
+	v.logger.Info("validatig " + pluginPath)
```
sbchaos, line 50: `validatig` -> `validating`

plugin/yaml/plugin.go
```diff
@@ -0,0 +1,135 @@
...
+)
+
+const (
+	Prefix = "optimus-plugin-"
+	Suffix = ".yaml"
+)
+
+type PluginSpec struct {
+	models.PluginInfoResponse    `yaml:",inline"`
+	models.GetQuestionsResponse  `yaml:",inline"` // PluginQuestion has extra attrs related to validation
```
smarch-int, line 23: remove the comment

models/plugin.go
```diff
@@ -355,17 +403,51 @@ type Plugin struct {
...
+func (s PluginList) Swap(i, j int) {
+	s[i], s[j] = s[j], s[i]
+}
+func (s PluginList) Less(i, j int) bool {
+	return s[i].Info().Name < s[j].Info().Name
 }
 
 func (p *Plugin) Info() *PluginInfoResponse {
-	resp, _ := p.Base.PluginInfo()
+	if p.Base != nil {
```
smarch-int, line 433: use `IsYamlPlugin()` instead

plugin/yaml/plugin.go
```diff
@@ -0,0 +1,135 @@
...
+	}, nil
+}
+
+func (PluginSpec) CompileAssets(_ context.Context, req models.CompileAssetsRequest) (*models.CompileAssetsResponse, error) { //nolint
+	return &models.CompileAssetsResponse{
+		Assets: req.Assets,
+	}, nil
+}
+
+func NewPlugin(pluginPath string) (models.CommandLineMod, error) {
```
smarch-int, line 94: rename to `NewPluginSpec` ?

cmd/plugin/validate.go
```diff
@@ -0,0 +1,108 @@
...
+
+func (v *validateCommand) RunE(_ *cobra.Command, _ []string) error {
+	fileInfo, err := os.Stat(v.path)
+	if err != nil {
+		return err
+	}
+	fm := fileInfo.Mode()
+	if fm.IsRegular() {
+		err := v.validateFile(v.path)
+		if err == nil {
```
sbchaos, line 99: Same function is handling error in 2 different ways can we use the standard notation of `err != nil`

smarch-int: which one seems better ?
```
err := v.validateFile(v.path)
if err != nil {
    return err
}
v.logger.Info("validation complete !")
return err
```
vs
```
err := v.validateFile(v.path)
if err == nil {
   v.logger.Info("validation complete !")
}
return err
```

mock/plugin.go
```diff
@@ -12,8 +12,8 @@ type SupportedPluginRepo struct {
 	mock.Mock
 }
 
-func (repo *SupportedPluginRepo) Add(plugin models.BasePlugin, mod models.CommandLineMod, mod2 models.DependencyResolverMod) error {
-	return repo.Called(plugin, mod, mod2).Error(0)
+func (repo *SupportedPluginRepo) Add(plugin models.BasePlugin, mod models.CommandLineMod, mod2 models.DependencyResolverMod, mod3 models.CommandLineMod) error {
```
sbchaos, line 15: Can we give better name to explain when we have 2 `models.CommandLineMod`

models/plugin.go
```diff
@@ -378,7 +460,7 @@ func (s *registeredPlugins) GetAll() []*Plugin {
 	for _, unit := range s.data {
 		list = append(list, unit)
 	}
-	return list
+	return s.sortPlugins(list)
```
sbchaos, line 463: Do we need to sort it every time, can we not sort it at creation time and use the list. ?

models/plugin_test.go
```diff
@@ -0,0 +1,164 @@
+package models
```
sbchaos, line 1: This should be in `models_test` package.

models/plugin.go
```diff
@@ -355,17 +403,51 @@ type Plugin struct {
 	// can be used in different circumstances
 	CLIMod        CommandLineMod
 	DependencyMod DependencyResolverMod
+	YamlMod       CommandLineMod
+}
+
+func (p *Plugin) IsYamlPlugin() bool {
+	return p.Base == nil
```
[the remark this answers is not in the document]
smarch-int: what's an invalid plugin ?

models/plugin.go
smarch-int: since repo is a map, either I can maintain sorted keys to return sorted list of plugins or sort every time version is being called (either the case it's once per invocation).
But I see that sort implementation is ugly here. I guess I can maintain sorted list of names. WDYT ?

plugin/yaml/plugin.go
```diff
@@ -0,0 +1,135 @@
...
+			Error:   err.Error(),
+		}, nil
+	}
+	return &models.ValidateQuestionResponse{
+		Success: true,
+	}, nil
+}
+
+func (p *PluginSpec) DefaultConfig(_ context.Context, req models.DefaultConfigRequest) (*models.DefaultConfigResponse, error) { //nolint
+	conf := []models.PluginConfig{}
```
sbchaos, line 64: Create empty value with var
eg `var conf []models.PluginConfig`

plugin/yaml/plugin.go
```diff
@@ -0,0 +1,135 @@
...
+
+	// adding defaultconfig (static, macros & referential config) from yaml
+	conf = append(conf, p.Config...)
+
+	return &models.DefaultConfigResponse{
+		Config: conf,
+	}, nil
+}
+
+func (p *PluginSpec) DefaultAssets(context.Context, models.DefaultAssetsRequest) (*models.DefaultAssetsResponse, error) { //nolint
```
sbchaos, line 82: Please specify the disabled linter with `//nolint` eg `// nolint:nilnil` currently it disables all the linters.

plugin/yaml/plugin.go
```diff
@@ -0,0 +1,135 @@
...
+			err = models.ErrNoSuchSpec
+		}
+		return nil, err
+	}
+	defer fd.Close()
+	pluginBytes, err := io.ReadAll(fd)
+	if err != nil {
+		return nil, err
+	}
+	plugin := PluginSpec{}
```
sbchaos, line 108: Use var to declare the variable.

https://github.com/uber-go/guide/blob/master/style.md#use-var-for-zero-value-structs

plugin/yaml/plugin.go
sbchaos: Name of function is `NewPlugin` or `NewPluginSpec` but it returns `models.CommandLineMod`

plugin/yaml/plugin.go
```diff
@@ -0,0 +1,135 @@
...
+func Init(pluginsRepo models.PluginRepository, discoveredYamlPlugins []string, pluginLogger hclog.Logger) {
+	for _, yamlPluginPath := range discoveredYamlPlugins {
+		yamlPlugin, err := NewPlugin(yamlPluginPath)
+		if err != nil {
+			pluginLogger.Error(fmt.Sprintf("plugin Init: %s", yamlPluginPath), err)
+			continue
+		}
+		pluginInfo, _ := yamlPlugin.PluginInfo()
+		if plugin, _ := pluginsRepo.GetByName(pluginInfo.Name); plugin != nil && !plugin.IsYamlPlugin() {
+			pluginLogger.Debug("skipping yaml plugin (as binary already added): ", pluginInfo.Name)
```
sbchaos, line 126: Should it be logged at Info level ?

plugin/yaml/plugin.go
```diff
@@ -0,0 +1,135 @@
...
+			continue
+		}
+		if err := pluginsRepo.Add(nil, nil, nil, yamlPlugin); err != nil {
```
sbchaos, line 129: This is not a good idea, can you create 2 separate add function for `AddBinary()` and `AddYaml()`, the current function does lot of things.

models/plugin.go
```diff
@@ -418,14 +502,24 @@ func (s *registeredPlugins) GetHooks() []*Plugin {
...
 	}
-	return list
+	return s.sortPlugins(list)
 }
 
-func (s *registeredPlugins) Add(baseMod BasePlugin, cliMod CommandLineMod, drMod DependencyResolverMod) error {
-	info, err := baseMod.PluginInfo()
-	if err != nil {
-		return err
+func (s *registeredPlugins) Add(baseMod BasePlugin, cliMod CommandLineMod, drMod DependencyResolverMod, yamlMod CommandLineMod) error {
```
sbchaos, line 508: This should be split in 2 separate method.

plugin/yaml/plugin.go
smarch-int: if you see the impl of `Add()`, you would notice `AddBinary()` and `AddYaml()` will be similar. 
Can add a wrapper to the existing `Add()` to be more explicit about binary or yaml though.
```
AddBinary() -> Add(base, cli, deps, nil)
AddYaml() -> Add(nil, nil, nil, yaml)
```
anyway, after migrating away from binary plugins,  the add signature will change.
adding new method now implicated more code change across (as change in interface, tests, mocks)

cmd/plugin/validate.go
```diff
@@ -46,13 +48,38 @@ func (v *validateCommand) PreRunE(_ *cobra.Command, _ []string) error {
...
-	v.logger.Info("validatig " + pluginPath)
+	v.logger.Info("\nValidating " + pluginPath)
 	if filepath.Ext(pluginPath) != ".yaml" {
 		return errors.New("expecting .yaml file at " + pluginPath)
 	}
-	plugin, err := yaml.NewPlugin(pluginPath)
-	v.logPluginAsYaml(&plugin)
+	var err error
+	var plugin models.CommandLineMod
+	if plugin, err = yaml.NewPluginSpec(pluginPath); err == nil {
```
sbchaos, line 75: Reduce nesting, check `err != nil` and return err in that case. Outdent rest of the block.

sbchaos, line 75: Reduce nesting, check `err != nil` and return err in that case. Outdent rest of the block.

cmd/version/version.go
```diff
@@ -115,7 +115,11 @@ func (v *versionCommand) printAllPluginInfos() {
 	v.logger.Info(fmt.Sprintf("\nDiscovered plugins: %d", len(plugins)))
 	for taskIdx, tasks := range plugins {
 		schema := tasks.Info()
-		v.logger.Info(fmt.Sprintf("\n%d. %s", taskIdx+1, schema.Name))
+		if tasks.YamlMod != nil {
+			v.logger.Info(fmt.Sprintf("\n%d. %s (%s)", taskIdx+1, schema.Name, "yaml"))
```
smarch-int, line 119: is this required ?

sbchaos (approved):

PR #454, opened 2022-07-21
refactor: simplify observer response for resource deployment

deryrahman:


core/sender/progress_count_deploy_resource.go
```diff
@@ -0,0 +1,26 @@
+package sender
+
+import (
+	pb "github.com/odpf/optimus/api/proto/odpf/optimus/core/v1beta1"
+)
+
+type deployResourceProgressCount struct {
```
sravankorumilli, line 7: We don't need the progress, as this is not helpful to the users and this response is tightly coupled with the optimus cli. So, better avoid it.

core/sender/log_status_deploy_resource.go
```diff
@@ -0,0 +1,20 @@
+package sender
+
+import (
+	pb "github.com/odpf/optimus/api/proto/odpf/optimus/core/v1beta1"
+)
+
+type deployResourceLogStatus struct {
```
sravankorumilli, line 7: Similar to job deploymentstatus we can do the same here

sravankorumilli (changes requested):

cmd/deploy/deploy.go
```diff
@@ -313,13 +313,11 @@ func (d *deployCommand) processResourceDeploymentResponse(
...
 ) error {
 	d.logger.Info("> Receiving responses:")
-	var counter int
 	spinner := progressbar.NewProgressBar()
 	defer spinner.Stop()
 
-	if !d.verbose {
-		spinner.StartProgress(totalSpecsCount, "please wait")
-	}
+	spinner.StartProgress(totalSpecsCount, "please wait")
```
sravankorumilli, line 319: we don't need this right?

cmd/deploy/deploy.go
```diff
@@ -328,19 +326,21 @@ func (d *deployCommand) processResourceDeploymentResponse(
 			}
 			return err
 		}
-		if resp.GetAck() {
-			if !resp.GetSuccess() {
-				d.logger.Error(resp.GetMessage())
+
+		if logStatus := resp.GetLogStatus(); logStatus != nil {
```
sravankorumilli, line 330: we can move the log handling to a single place, this is duplicate code in all the commands

sravankorumilli (changes requested):

cmd/deploy/deploy.go
deryrahman: yes, we don't need it anymore. deleted

cmd/deploy/deploy.go
deryrahman: let's doing it once it's merged on `simplify-observer-response` branch

PR #472, opened 2022-07-27
fix: temp convert resource managers from array to single value

SumitAgrawal03071989:
This is temprory arrangement to be able to specify resource manager config as environment variable.


sravankorumilli:
this may not be needed now right @SumitAgrawal03071989

SumitAgrawal03071989:
Yes, closing the PR

Issue #490, opened 2022-08-02
Preserve history of Airflow sensor task by having the same sensor name as internal sensor task


arinda-arif:
There are 2 kinds of Optimus sensor: internal and external optimus sensor. Currently, internal sensor uses the job name as the sensor ID, while external sensor uses host, project, and the job name.

This might causes an issue when a job moved from 1 server to another server, where the history of the sensor will not be preserved as the task ID has changed.

sravankorumilli:
this is already done

PR #491, opened 2022-08-02
refactor: change external optimus sensor task name similar with internal sensor name

arinda-arif:
Resolving https://github.com/odpf/optimus/issues/490


SumitAgrawal03071989 (approved):

irainia (approved):

ext/scheduler/airflow2/resources/base_dag.py
```diff
@@ -227,16 +227,15 @@
 {{- end}}
 
 {{- range $_, $dependency := $.Job.ExternalDependencies.OptimusDependencies}}
-{{ $identity := print $dependency.Name "-" $dependency.ProjectName "-" $dependency.JobName }}
-wait_{{$identity | replace "-" "__dash__" | replace "." "__dot__"}} = SuperExternalTaskSensor(
+wait_{{$dependency.JobName | replace "-" "__dash__" | replace "." "__dot__"}} = SuperExternalTaskSensor(
```
sravankorumilli, line 230: i believe we can do this change only on the id's the name of the sensor can still be the same.

sravankorumilli: we can set the name too such that it is evident to users that it is external dependency

sravankorumilli (changes requested):

sravankorumilli (approved):

Issue #499, opened 2022-08-08
Bug external dependencies error breaks deployment


irainia:
**Describe the bug**
During deployment process, when we want to get external dependencies, we faced an error. This error breaks or stops the whole deployment process.

**To Reproduce**
1. create a job that depends on external Optimus
2. ensure the origin Optimus cannot access targetted Optimus
3. run deployment process

**Expected behavior**
* for external inferred dependencies, if there's any error, just skip it
* for external static dependencies, if there's any error:
  * we store it for logging,
  * we still continue to the next job, and
  * we fail the deployment while providing the log

PR #500, opened 2022-08-08
fix: bug breaking deploy on resource manager failure

irainia:
This is to address [this issue](https://github.com/odpf/optimus/issues/499).

### DIFF
--- a/job/external_dependency_resolver.go
+++ b/job/external_dependency_resolver.go
@@ -45,10 +45,7 @@ func (e *externalDependencyResolver) FetchInferredExternalDependenciesPerJobName
 	}
 	externalDependencyPerJobName := make(map[string]models.ExternalDependency)
 	for jobName, filters := range unresolvedDependenciesPerJobName {
-		optimusDependencies, err := e.fetchInferredOptimusDependencies(ctx, filters)
-		if err != nil {
-			return nil, err
-		}
+		optimusDependencies := e.fetchInferredOptimusDependencies(ctx, filters)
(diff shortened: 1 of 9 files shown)


job/external_dependency_resolver.go
```diff
@@ -65,11 +62,8 @@ func (e *externalDependencyResolver) FetchStaticExternalDependenciesPerJobName(c
 	var unknownDependencies []models.UnknownDependency
 	externalDependencyPerJobName := make(map[string]models.ExternalDependency)
 	for jobName, toBeResolvedDependency := range unresolvedDependenciesPerJobName {
-		optimusDependencies, unresolvedFromExternal, err := e.fetchStaticOptimusDependencies(ctx, toBeResolvedDependency)
-		if err != nil {
-			return nil, nil, err
-		}
-		unknownDependencies = convertUnresolvedToUnknownDependencies(jobName, unresolvedFromExternal)
+		optimusDependencies, unresolvedFromExternal := e.fetchStaticOptimusDependencies(ctx, toBeResolvedDependency)
+		unknownDependencies = e.convertUnresolvedToUnknownDependencies(jobName, unresolvedFromExternal)
```
arinda-arif, line 66: @irainia let's append this `unknownDependencies` so it will not be replaced in every loop? looks like this is missed in the previous commit.

arinda-arif: apart from this, looks good.

irainia: updated

arinda-arif (approved):

PR #529, opened 2022-08-18
feat: call job run input to compile + fetch assets

deryrahman:
Command job run-input will fail if the task instance before `job_start_event` feat introduced is triggered. It happens because the old instance is not registered in the `job_run` table yet. So we need to register the instance either using `admin build instance` or manually trigger `job_start_event`.

In this implementation, fail to run command `job run-input` will fallback to `admin build instance`

### DIFF
--- a/entrypoint_init_container.sh
+++ b/entrypoint_init_container.sh
@@ -18,7 +18,16 @@ echo "OPTIMUS_HOST:$OPTIMUS_HOST"
 echo ""
 
 echo "-- initializing optimus assets"
-OPTIMUS_ADMIN_ENABLED=1 optimus admin build instance "$JOB_NAME" --project-name \
+OPTIMUS_ADMIN_ENABLED=1 optimus job run-input "$JOB_NAME" --project-name \
 	"$PROJECT" --output-dir "$JOB_DIR" \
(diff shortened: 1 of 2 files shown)


entrypoint_init_container.sh
```diff
@@ -18,7 +18,16 @@ echo "OPTIMUS_HOST:$OPTIMUS_HOST"
...
 echo "-- initializing optimus assets"
-OPTIMUS_ADMIN_ENABLED=1 optimus admin build instance "$JOB_NAME" --project-name \
+OPTIMUS_ADMIN_ENABLED=1 optimus job run-input "$JOB_NAME" --project-name \
 	"$PROJECT" --output-dir "$JOB_DIR" \
 	--type "$INSTANCE_TYPE" --name "$INSTANCE_NAME" \
-	--scheduled-at "$SCHEDULED_AT" --host "$OPTIMUS_HOST"
\ No newline at end of file
+	--scheduled-at "$SCHEDULED_AT" --host "$OPTIMUS_HOST"
+
+if [ $? -ne 0 ]; then
```
smarch-int, line 26: why fallback ?

smarch-int (approved):
LGTM!

Mryashbhardwaj (approved):

PR #530, opened 2022-08-18
feat : cli syncs yaml plugins from server

smarch-int:
Changes:
- during plugin installation, server creates a zip file (`yaml-plugins.zip`) containing all yaml plugins \
   and cli downloads the zip from server
- server : `http://{{host}}/plugins`
- client : `optimus plugin sync -c optimus.yaml`
- refactor : move plugin logic from cmd to plugin pkg.
- plugin discovery is static -  `.plugins` (for client and server)
- remove plugins.Dir from server plugin config   (to keep it simple)

Implementation Details:
```go
type IPluginManager interface {
	Install(dst string, sources ...string) error
	Archive(name string) error        // by server
	UnArchive(src, dest string) error // by client
}
```


plugin/plugin_manager.go
```diff
@@ -0,0 +1,189 @@
...
+	if err := p.client.Get(); err != nil {
+		p.logger.Error(fmt.Sprintf("Error installing plugin from [%s]", src))
+		return err
+	}
+	p.logger.Info(fmt.Sprintf("Success installing plugin from [%s]", src))
+	return nil
+}
+
+func (p *PluginManager) Install(dst string, sources ...string) error {
+	// p.preCleanUp(dst) -- TODO: after making plugin discovery static
```
smarch-int, line 73: remove comment

cmd/plugin/install.go
```diff
@@ -6,17 +6,19 @@ import (
...
+	serverConfig   *config.ServerConfig
+	configFilePath string `default:"config.yaml"`
 }
 
 // NewInstallCommand initializes plugin install command
-func NewInstallCommand(serverConfig *config.ServerConfig) *cobra.Command {
+func NewInstallCommand() *cobra.Command {
 	install := &installCommand{
-		serverConfig: serverConfig,
+		serverConfig: &config.ServerConfig{},
```
sbchaos, line 21: This should use the default value for pointer, we are overriding this value in preRun.

cmd/plugin/install.go
```diff
@@ -25,22 +27,20 @@ func NewInstallCommand(serverConfig *config.ServerConfig) *cobra.Command {
...
 	return cmd
 }
 
 func (i *installCommand) PreRunE(_ *cobra.Command, _ []string) error {
-	i.logger = logger.NewClientLogger(i.serverConfig.Log)
+	c, err := config.LoadServerConfig(i.configFilePath)
+	if err != nil {
+		return err
+	}
+	*i.serverConfig = *c
```
sbchaos, line 39: Why are we using pointer to a pointer for setting value. ?
LoadServerConfig returns a pointer to `ServerConfig` which we can set directly in struct.

```go
	i.serverConfig = c
```

cmd/plugin/sync.go
```diff
@@ -0,0 +1,77 @@
...
+	"github.com/spf13/cobra"
+
+	"github.com/odpf/optimus/config"
+	"github.com/odpf/optimus/plugin"
+)
+
+// NewInstallCommand initializes plugin install command
+func NewSyncCommand() *cobra.Command {
+	sync := &syncCommand{
+		clientConfig: &config.ClientConfig{},
```
sbchaos, line 19: Can we use default value for pointer here ?

cmd/plugin/sync.go
```diff
@@ -0,0 +1,77 @@
...
+	configFilePath string
+	clientConfig   *config.ClientConfig
+}
+
+func (s *syncCommand) PreRunE(_ *cobra.Command, _ []string) error {
+	c, err := config.LoadClientConfig(s.configFilePath)
+	if err != nil {
+		return err
+	}
+	*s.clientConfig = *c
```
sbchaos, line 42: Can be written as 

```go
	i.serverConfig = c
```

plugin/plugin_manager.go
```diff
@@ -0,0 +1,189 @@
...
+var (
+	// for plugin installation, discovery & sync on both client and server
+	PluginsDir         = ".plugins"
+	PluginsArchiveName = "yaml-plugins.zip"
+)
+
+type IPluginManager interface {
+	Install(dst string, sources ...string) error
+	Archive(name string) error        // by server
+	UnArchive(src, dest string) error // by client
```
sbchaos, line 28: If the interface is not consumed as a whole, should be broken down in 2 different interfaces and relevant modules should depend on the corresponding one.

plugin/plugin_manager.go
```diff
@@ -0,0 +1,189 @@
...
+}
+
+func NewPluginManager() IPluginManager {
```
sbchaos, line 31: Constructor should return struct.

plugin/plugin_manager.go
smarch-int: This interface is what the plugin manager library is exposing. the client that uses this manager can have restricted interfaces. For now I am keeping things simple.

cmd/plugin/sync.go
```diff
@@ -0,0 +1,75 @@
...
+	c, err := config.LoadClientConfig(s.configFilePath)
+	if err != nil {
+		return err
+	}
+	s.clientConfig = c
+	return nil
+}
+
+func (s *syncCommand) downloadArchiveFromServer() error {
+	optimusServerURL := fmt.Sprintf("http://%s/%s", s.clientConfig.Host, "plugins") // will url work ?
```
smarch-int, line 45: @sbchaos can we rely on this url format ?

plugin/plugin_manager.go
```diff
@@ -0,0 +1,188 @@
...
+		if err := p.installOne(dst, src); err != nil {
+			p.logger.Error("*** Plugin Installation Aborted !!. Please check if plugin.artifacts are correct")
+			return err
+		}
+	}
+	p.logger.Info(fmt.Sprintf("Success installing plugin in dir=%s", dst))
+	return nil
+}
+
+func SanitizeArchivePath(d, t string) (v string, err error) {
```
sbchaos, line 83: Why are we using this function outside this module ? if not it should start with lowercase letter

cmd/plugin/sync.go
sbchaos: Can the host include the scheme ?  if we need to support https or ftp path, this will be restrictive.
Just append path `plugins` to the Host if we can accept it along with scheme.

sbchaos (approved):

PR #536, opened 2022-08-23
refactor: command package

deryrahman:


deryrahman:
The `server` package is moved to root instead of `cmd/internal`. The reason is, server has strong dependencies with packages in root. And, as per my understanding, `internal` package should only have low dependencies with others. Here's why package `logger`, `connectivity`, etc can reside in `internal` package.

Optimus folder structure as per commit [4b888c3f45828d5cd502ab07428da72db57b9830](https://github.com/odpf/optimus/tree/4b888c3f45828d5cd502ab07428da72db57b9830)
```
.
├── api *
├── cmd
├── compiler *
├── config
├── core *
├── datastore *
├── dist
├── docs
├── ext *
├── extension
├── grafana_dashboards
├── job *
├── mock
├── models *
├── plugin *
├── scripts
├── service *
├── store *
├── tests
└── utils
```

Marked packages (*) are belong to `server`

irainia (approved):

Issue #531, opened 2022-08-19
Refactor `createJobSpecification` api


Mryashbhardwaj:
**Description**

- Currently `createJobSpecification`  only accepts one job spec at a time for deployment.
- we have 2 similar implementations for creating and deploying a job, in `CreateJobSpecification` vs `DeployJobSpecification` we aim to reuse the code flow of `DeployJobSpecification`

**Acceptance Criteria**

- [ ] create api should be able to accept one or more JobSpec for creation and deployment
- [ ] deprecate sync function from the createJobSpecification flow

**Out of Scope**

1. Renaming the API
2. `CreateJobSpecification` will remain as non stream api, and will do the deployment async(non-blocking). it will return deployment id for the consumer to be able to poll the deployment status.

PR #537, opened 2022-08-24
feat: add AddJobSpecifications api & refactor createJobSpecification api

Mryashbhardwaj:
fixes:  https://github.com/odpf/optimus/issues/531 

Proton PR: https://github.com/odpf/proton/pull/179 

PS: 

- added a new api in the name `AddJobSpecifications` to fix  https://github.com/odpf/optimus/issues/531 
- `createJobSpecification` now schedules deployment async , and returns deploymentId in success msg


api/handler/v1beta1/job_spec.go
```diff
@@ -163,30 +163,44 @@ func (sv *JobSpecServiceServer) CreateJobSpecification(ctx context.Context, req
...
+	}
+
+	jobNames := make([]string, len(jobSpecs))
+	for i, jobSpec := range jobSpecs {
+		jobNames[i] = jobSpec.Name
 	}
 
 	// validate job spec
-	if err = sv.jobSvc.Check(ctx, namespaceSpec, []models.JobSpec{jobSpec}, logWriter); err != nil {
+	if err = sv.jobSvc.Check(ctx, namespaceSpec, jobSpecs, logWriter); err != nil {
```
arinda-arif, line 182: How about call to the JobService only once, so we will put the flow of having to do checking, saving, and deployment scheduling in JobService?
- We don't need to export IdentifyAndPersistJobSources, BulkCreate & ScheduleDeployment
- Since JobService.Create already exist, we need to think what should be the proper name of this new function & the existing Create function

job/service.go
```diff
@@ -393,96 +393,6 @@ func (srv *Service) bulkDelete(ctx context.Context, namespace models.NamespaceSp
 	}
 }
 
-// Sync fetches all the jobs that belong to a project, resolves its dependencies
```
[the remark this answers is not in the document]
arinda-arif: Second thought, looks like for dependency_resolver it will still be needed for Backup & Replay, but can you check the airflow one?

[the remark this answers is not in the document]
Mryashbhardwaj: i checked the flow of Sync , most the underlying function where used in other flows. The one's, that were not being used have been removed,

Btw, is there any specific function that you are thinking about.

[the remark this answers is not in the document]
arinda-arif: It is the `DeployJobs` in `ext/scheduler/airflow2/airflow.go`. However, turns out it is used by ManualScheduler in the `JobService.Run`, so would also need to understand this usage. We can have a discussion regarding this.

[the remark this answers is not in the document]
sravankorumilli: I believe it is better to clean up the prime scheduler too. Lets review the code around it & we can do the cleanup accordingly.

models/job.go
```diff
@@ -419,6 +418,9 @@ type HTTPDependency struct {
 type JobService interface {
 	// Create constructs a Job and commits it to a storage
 	Create(context.Context, NamespaceSpec, JobSpec) (JobSpec, error)
+
+	CreateAndDeployJobSpecifications(context.Context, NamespaceSpec, []JobSpec, writer.LogWriter) (DeploymentID, error)
```
arinda-arif, line 422: @Mryashbhardwaj should we just keep the name `CreateAndDeploy`? Since for refresh and deploy we named it `Refresh` and `Deploy`.

Mryashbhardwaj: yeah we should do that, ill fix this

models/scheduler.go
```diff
@@ -25,8 +25,8 @@ type SchedulerUnit interface {
 	GetName() string
 	VerifyJob(ctx context.Context, namespace NamespaceSpec, job JobSpec) error
 	ListJobs(ctx context.Context, namespace NamespaceSpec, opts SchedulerListOptions) ([]Job, error)
-	DeployJobs(ctx context.Context, namespace NamespaceSpec, jobs []JobSpec, obs progress.Observer) error
-	DeployJobsVerbose(ctx context.Context, namespace NamespaceSpec, jobs []JobSpec) (JobDeploymentDetail, error)
+	// depricated for deploy in optimus
```
arinda-arif, line 28: @Mryashbhardwaj you might want to recheck this comment

models/job.go
Mryashbhardwaj: done

models/scheduler.go
Mryashbhardwaj: done

api/handler/v1beta1/job_spec.go
```diff
@@ -163,30 +163,31 @@ func (sv *JobSpecServiceServer) CreateJobSpecification(ctx context.Context, req
...
 	}
 
-	// validate job spec
-	if err = sv.jobSvc.Check(ctx, namespaceSpec, []models.JobSpec{jobSpec}, logWriter); err != nil {
-		return nil, status.Errorf(codes.Internal, "spec validation failed\n%s", err.Error())
+	jobNames := make([]string, len(jobSpecs))
+	for i, jobSpec := range jobSpecs {
+		jobNames[i] = jobSpec.Name
 	}
+	deployID, err := sv.jobSvc.CreateAndDeploy(ctx, namespaceSpec, jobSpecs, logWriter)
```
arinda-arif, line 180: let's remove the space just below this line, so there should be no gap between this and the error check. also, you might want to add the failure case unit test for `CreateAndDeploy` to increase the coverage.

arinda-arif: if you can also add a unit test for deserializing job failure in this function, that would be great.

arinda-arif:
The changes look good. After fixing the coverage and changing the proton commit to the merged one, we can merge this.

api/handler/v1beta1/job_spec.go
Mryashbhardwaj: nice idea, added test cases

api/handler/v1beta1/job_spec.go
```diff
@@ -163,30 +163,57 @@ func (sv *JobSpecServiceServer) CreateJobSpecification(ctx context.Context, req
...
+	}
+
+	jobNames := make([]string, len(jobSpecs))
+	for i, jobSpec := range jobSpecs {
+		jobNames[i] = jobSpec.Name
 	}
+	deployID, err := sv.jobSvc.CreateAndDeploy(ctx, namespaceSpec, jobSpecs, logWriter)
 
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "jobs %s is/are created but deployment scheduling failed for project %s, error:: %s", strings.Join(jobNames, ", "), req.GetProjectName(), err.Error())
```
irainia, line 210: I believe it will be too verbose in the log to put all of the job names. How about we limit it? For example, let's say we have 20 jobs being added, then we can have:
```
jobs [jobA, jobB] and 18 others are created but deployment ...
```

irainia (changes requested):

api/handler/v1beta1/job_spec.go
Mryashbhardwaj: Sure we can do that, 
but i thought we will be using this end point for deploying a small (2, 3 maybe 5 ) number  of jobs only.

sravankorumilli: we don't need to mention the job names if we don't have the idea of which jobs it failed. We can just log the error. Otherwise it will give a misinterpretation that creation of all jobs failed.

Mryashbhardwaj: actually right, fixing that

irainia (approved):

api/handler/v1beta1/job_spec.go
```diff
@@ -163,30 +163,53 @@ func (sv *JobSpecServiceServer) CreateJobSpecification(ctx context.Context, req
...
+	deployID, err := sv.jobSvc.CreateAndDeploy(ctx, namespaceSpec, jobSpecs, logWriter)
+
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "job addition failed for project %s, error:: %s", req.GetProjectName(), err.Error())
 	}
+	runtimeDeployJobSpecificationCounter.Inc()
+
+	return &pb.CreateJobSpecificationResponse{
+		Success: true,
+		Message: fmt.Sprintf("job is created and queued for deployment on project %s, with DeploymentId :: %s", req.GetProjectName(), deployID.UUID().String()),
```
arinda-arif, line 182: - looks like the colon is doubled
- lets change `DeploymentID` to `Deployment ID`

api/handler/v1beta1/job_spec.go
```diff
@@ -163,30 +163,53 @@ func (sv *JobSpecServiceServer) CreateJobSpecification(ctx context.Context, req
...
+	for _, spec := range reqJobSpecs {
+		jobSpec, err := FromJobProto(spec, sv.pluginRepo)
+		if err != nil {
+			return nil, status.Errorf(codes.Internal, "cannot deserialize job: \n%s", err.Error())
+		}
+		jobSpecs = append(jobSpecs, jobSpec)
 	}
 
+	deployID, err := sv.jobSvc.CreateAndDeploy(ctx, namespaceSpec, jobSpecs, logWriter)
+
```
arinda-arif, line 204: let's remove this extra linespace before error check

arinda-arif (approved):

Issue #417, opened 2022-06-23
Add Proper Migration Up and Down


irainia:
### Background

If we look at the latest commit (referring to [this](79fdefa1012caed723dd681e3e02eff79830f9e7)), internally we will have migration mechanism being used. The migration `up` is executed whenever we run the server. However, if we check even further, no `down` migration is provided. So, even if there's an issue with the current database schema, it's quite tricky to roll back.

### Proposal

To address this, this issue proposes to have such functionality being provided in the Optimus. At the high level, more or less, it will be like the following:

```zsh
optimus server migrate up
# it will execute all migrate `up`

optimus server migrate down
# it will execute all migrate `down`

optimus server migrate {n}
# n is integer number, with positive means `up` n-time and negative means `down` n-time
```

The command or the mechanism is flexible, but the point is that the `up` and `down` are both defined properly.

### Additional Context

Since the mentioned commit use [golang-migrate/migrate](github.com/golang-migrate/migrate/v4), we can use `.Steps(int)` method for the `n` up or down times.

sbchaos:
We also need to update the version of [golang-migrate](https://github.com/odpf/optimus/issues/github.com/golang-migrate/migrate/v4) library to latest version in a patch release on top of current release, so that later if we need to do a rollback, we will not be stuck in incompatible version.

irainia:
### TLDR

* we will have Optimus migration up and rollback only
* migration up is run automatically when server up
* migration rollback is run using client command, for example:

```zsh
optimus migration rollback
```

* before rollback, ensure no optimus server is running

### Approach

This analysis is based on Optimus v0.3.5 with migration version 27.

#### Functonalities

The following is the available functionalities for migration:

* migration up: migrate to the latest
  
  this function is run when we start optimus server. it means whenever there's a new release, we will migrate the latest migration version. for this case, we assume that the current optimus server that does the migration up is the latest version. meaning, the database migration version should be less than or equal to optimus server's migration version.

  the following is the flow:
  1. get the previous optimus version based on the currently active migration version
  1. we migrate all the way to the latest version
  1. create (if not exist) a table to store:
     * current optimus server version
     * optimus server version prior to the current version (used when rollback)
     * current optimus server's migration version
  1. get the current optimus server version
  1. get the current mgiration version
  1. store current and previous (if exist) optimus server versions and the current migration server to the table

[...]

PR #538, opened 2022-08-25
feat: add migration implementation

irainia:
This PR is to address [this issue](https://github.com/odpf/optimus/issues/417). In the implementation, there are some minor differences compared to the mentioned approach.

Differences:
* some additional checks are done, such as the version of Optimus that does migration with the currently active one
* at some states, migration can be skipped, such as when there's no changes on Optimus version
* on migration Rollback, if the previous version is not found, it's not returning error, but only a warning
* changes the table that will store the migration version from `migration_versions` into `migration_steps` for more clarity and to avoid conflict name during its implementation code
* add record removal for `migration_steps` when migration Rollback is successful

*Note: additional detail may be added later*


store/postgres/migration.go
```diff
@@ -0,0 +1,185 @@
...
+		return nil, errors.New("incoming optimus version is empty")
+	}
+	if dbConnURL == "" {
+		return nil, errors.New("database connection url is empty")
+	}
+	m, err := newMigrate(dbConnURL)
+	if err != nil {
+		return nil, fmt.Errorf("error initializing migrate: %w", err)
+	}
+	db, err := gorm.Open(postgres.Open(dbConnURL))
```
sbchaos, line 52: We are opening a separate connection for migrate, where are we closing this connection ?

irainia: updated by adding explicit close statement

store/postgres/migration.go
```diff
@@ -0,0 +1,232 @@
...
+	}
+	return m.removeMigrationStep(ctx, dbClient, latestStep)
+}
+
+func (m *migration) newMigrationClient() (migrationClient *migrate.Migrate, cleanup func(), err error) {
+	path := "migrations"
+	sourceDriver, sourceDriverErr := iofs.New(migrationFs, path)
+	if sourceDriverErr != nil {
+		err = fmt.Errorf("error initializing source driver: %w", sourceDriverErr)
+		return
```
sbchaos, line 146: Do not use naked return, define values in function.

sbchaos, line 146: Avoid naked return, multiple return values are hard to keep in head.
Also assigning to `err` in case of actual error.

store/postgres/migration.go
```diff
@@ -0,0 +1,232 @@
...
+		}
+	}
+	return
+}
+
+func (m *migration) newDBClient() (dbClient *gorm.DB, cleanup func(), err error) {
+	gormDB, gormDBErr := gorm.Open(postgres.Open(m.dbConnURL))
+	if gormDBErr != nil {
+		err = fmt.Errorf("error initializing gorm db: %w", gormDBErr)
+		return
```
sbchaos, line 171: Use explicit returns

store/postgres/migration.go
irainia: since we have more than two returns, using named returns is so that the caller could understand in a glance what each return is about. in some cases, they are good to implement. however, I do agree that in the current case, where the function is a bit long, it's not a good approach. I already updated it.

store/postgres/migration.go
irainia: updated

sbchaos (approved):

Issue #494, opened 2022-08-02
Provide Upstream job runs and downstream jobs as part of `Optimus job inspect` command 


Mryashbhardwaj:
**Description**
Provide upstream job runs that the subject job will depend for a given schedule date.

**Acceptance Criteria**

- [ ] Print List of Upstream job runs for all the dependencies of the Subject job for given schedule times

Depends on : https://github.com/odpf/optimus/issues/492

**Tech Details**
Post analysis Capture the details of the work needed to finish the task.

- [ ] Use existing Window logic and sensor logic to compute the required schedule times of upstream jobs.

Mryashbhardwaj:
duplicate of https://github.com/odpf/optimus/issues/576

Issue #493, opened 2022-08-02
download rendered job assets via `Optimus job inspect` command


Mryashbhardwaj:
**Description**
Building upon https://github.com/odpf/optimus/issues/492 , add support for downloading rendered job assets.

**Acceptance Criteria**

- [ ] Download job assets with support for custom schedule time in specified local directory 

Depends on : https://github.com/odpf/optimus/issues/492

**Tech Details**
Post analysis Capture the details of the work needed to finish the task.

- [ ] Take inspiration from Optimus job render command

Mryashbhardwaj:
handled in https://github.com/odpf/optimus/issues/577

Issue #492, opened 2022-08-02
Provide `Optimus Job inspect` Command


Mryashbhardwaj:
**Description**
Provide a command which can inspect an Optimus job using its dependencies, while running certain validations too.

**Acceptance Criteria**
The command should be able to 
- [x] Query syntax Checks.
- [ ] ~perform Destination and source table validations & show errors~
- [x] give Soft warnings
    - [x]  if Enabled Catch Up ? 
    - [x]  If the subject job shares a destination table with an existing job 
- [x] This should work for local job spec & the job in the server
    - [x]  User should be provided an option for using local vs server for cli users. 
- [x] Job Add & Deploy Flows should be integrated with these hard validations.
    

**Out of Scope**
1. Job runs details 
2. Dependency jobs data quality metric
3. compiled assets 
4. perform Destination and source table validations & show errors
   1. this needs to be handled on plugins
   2. Optimus Is unaware of the jobSpecification defined by plugins{BQ2BQ, GCS2BQ} 
   3. Based on User Specification what are the Sources? Destination? information current design a plugin is responsible for sharing this information.

**Tech Details**
- color scheme 
    - hard warning - red
    - soft - yellow
    - information - white
- if the job already exists in server it should override that job and handle this.

PR #543, opened 2022-08-29
feat: Job Inspect

Mryashbhardwaj:
fixes : #494  #492 #493


job/service.go
```diff
@@ -306,6 +310,31 @@ func (srv *Service) Check(ctx context.Context, namespace models.NamespaceSpec, j
...
+		mullErr = multierror.Append(mullErr, fmt.Errorf("failed to generate dependencies::%w", err))
+	}
+	if deps != nil {
+		dependencies = deps.Dependencies
+	}
+
+	return destination, dependencies, mullErr
+}
+
 func (srv *Service) GetTaskDependencies(ctx context.Context, namespace models.NamespaceSpec, jobSpec models.JobSpec) (models.JobSpecTaskDestination,
```
Mryashbhardwaj, line 338: @sbchaos , is it okay if instead of returning `error` , we return ,`multierror.Error` .

if we can do that then we can avoid the need of adding a new function `GetJobSourceAndDestination`

client/cmd/job/inspect.go
```diff
@@ -0,0 +1,206 @@
...
+}
+
+// NewInspectCommand initializes command for inspecting job specification
+func NewInspectCommand() *cobra.Command {
+	inspect := &inspectCommand{
+		clientConfig: &config.ClientConfig{},
+	}
+	cmd := &cobra.Command{
+		Use:     "inspect",
+		Short:   "inspect optimus job specification using local and server spec",
```
arinda-arif, line 50: let's use capital for the short & long command description's first letter

client/cmd/job/inspect.go
```diff
@@ -0,0 +1,206 @@
...
+		Long:    "inspect optimus job specification using local and server spec",
+		Example: "optimus job inspect [<job_name>] [--server]",
+		Args:    cobra.MinimumNArgs(1),
+		RunE:    inspect.RunE,
+		PreRunE: inspect.PreRunE,
+	}
+	cmd.Flags().StringVarP(&inspect.configFilePath, "config", "c", config.EmptyPath, "File path for client configuration")
+	cmd.Flags().StringVar(&inspect.host, "host", "", "Optimus service endpoint url")
+	cmd.Flags().Bool(optimusServerFetchFlag, false, "fetch jobSpec from server")
```
arinda-arif, line 59: also first-letter capital on this boolean flag's usage description

internal/store/postgres/job_spec_repository.go
```diff
@@ -132,7 +132,7 @@ func (repo jobSpecRepository) GetStaticDependenciesPerJobID(ctx context.Context,
 	return repo.adapter.groupToDependenciesPerJobID(jobDependencies)
 }
 
-func (repo jobSpecRepository) getDependentJobsInferred(ctx context.Context, jobSpec *models.JobSpec) ([]models.JobSpec, error) {
+func (repo jobSpecRepository) GetDependentJobsInferred(ctx context.Context, jobSpec *models.JobSpec) ([]models.JobSpec, error) {
```
arinda-arif, line 135: looks like this does not need to be exported now?

job/service.go
```diff
@@ -303,6 +307,46 @@ func (srv *Service) Check(ctx context.Context, namespace models.NamespaceSpec, j
 	return err
 }
 
+func (srv *Service) GetJobBasicInfo(ctx context.Context, jobSpec models.JobSpec) models.JobBasicInfo {
+	var jobBasicInfo models.JobBasicInfo
+	jobBasicInfo.Spec = jobSpec
+
+	dest, err := srv.pluginService.GenerateDestination(ctx, jobSpec, jobSpec.NamespaceSpec)
+	if err != nil {
+		jobBasicInfo.Log.Write(writer.LogLevelError, fmt.Sprintf("unable to generate destination, err:%v", err))
```
arinda-arif, line 316: let's add space after colon

job/service.go
```diff
@@ -303,6 +307,46 @@ func (srv *Service) Check(ctx context.Context, namespace models.NamespaceSpec, j
...
+	}
+
+	deps, err := srv.pluginService.GenerateDependencies(ctx, jobSpec, jobSpec.NamespaceSpec, false)
+	if err != nil {
+		jobBasicInfo.Log.Write(writer.LogLevelError, fmt.Sprintf("failed to generate dependencies, err:%v", err))
+	} else {
+		if deps != nil {
+			jobBasicInfo.JobSource = deps.Dependencies
+		} else {
+			jobBasicInfo.Log.Write(writer.LogLevelWarning, "no dependencies detected")
```
arinda-arif, line 333: should we change the error message to `no job sources detected`? not finding anything does not mean there are no dependencies for this job, there might be static dependencies.

arinda-arif: we might also need to change `failed to generate dependencies` to `failed to generate job sources`, to avoid confusion with job dependency/upstream section later

job/service.go
```diff
@@ -303,6 +307,46 @@ func (srv *Service) Check(ctx context.Context, namespace models.NamespaceSpec, j
...
+		}
+	}
+	srv.Check(ctx, jobSpec.NamespaceSpec, []models.JobSpec{jobSpec}, &jobBasicInfo.Log)
+
+	if jobSpec.Behavior.CatchUp {
+		jobBasicInfo.Log.Write(writer.LogLevelWarning, "Catchup is enabled")
+	}
+	if dupDestJobName, err := srv.IsJobDestinationDuplicate(ctx, jobSpec); err != nil {
+		jobBasicInfo.Log.Write(writer.LogLevelError, " could not perform duplicate job destination check, err:"+err.Error())
```
arinda-arif, line 342: let's keep the warn/error log similar? seeing some with extra space in the front, and some using a capital letter.

job/service.go
```diff
@@ -760,6 +804,21 @@ func (srv *Service) fetchSpecsForGivenJobNames(ctx context.Context, projectSpec
 	return jobSpecs, nil
 }
 
+func (srv *Service) IsJobDestinationDuplicate(ctx context.Context, jobSpec models.JobSpec) (string, error) {
+	jobWithSameDestination, err := srv.jobSpecRepository.GetJobByResourceDestination(ctx, jobSpec.ResourceDestination)
```
arinda-arif, line 808: looks like this will be useful if the user is using local job spec? since if using server job spec, it will always return the first record encountered in db, and will not detect the existing duplicated job destination in the system.

job/service_test.go
```diff
@@ -2120,4 +2120,68 @@ func TestService(t *testing.T) {
...
+			pluginService.On("GenerateDependencies", ctx, jobSpec, namespaceSpec, true).Return(&models.GenerateDependenciesResponse{Dependencies: []string{"dependency1urn"}}, nil)
+			pluginService.On("GenerateDependencies", ctx, jobSpec, namespaceSpec, false).Return(&models.GenerateDependenciesResponse{Dependencies: []string{"dependency1urn"}}, nil)
+
+			batchScheduler := new(mock.Scheduler)
+			defer batchScheduler.AssertExpectations(t)
+			batchScheduler.On("VerifyJob", ctx, namespaceSpec, jobSpec).Return(nil)
+
+			jobSpecRepo := new(mock.JobSpecRepository)
+			defer jobSpecRepo.AssertExpectations(t)
+			//jobSpecRepo.On("GetJobByName", ctx, jobSpec.GetName()).Return([]models.JobSpec{jobSpec}, nil)
```
arinda-arif, line 2164: lets remove this comment if not needed

internal/lib/cron/cron.go
```diff
@@ -63,3 +63,15 @@ func (s *ScheduleSpec) getEarliestTimeToStartCron(currTime time.Time) time.Time
 	}
 	return startTime
 }
+
+func (s *ScheduleSpec) GetExpectedRuns(startTime, endTime time.Time) []time.Time {
```
arinda-arif, line 67: is this included in the scope of this PR?

client/cmd/internal/logger/color.go
```diff
@@ -0,0 +1,35 @@
...
+	ColoredNotice  = fmt.Sprintf
+	ColoredError   = fmt.Sprintf
+
+	tp          = termenv.EnvColorProfile()
+	ColorYellow = tp.Color("#FFAF00")
+	ColorWhite  = tp.Color("#FFFFFF")
+	ColorRed    = tp.Color("#D70000")
+)
+
+func InitializeColor() {
```
arinda-arif, line 22: is it intended to introduce this and not using the existing color function that being used in other commands?

job/service.go
Mryashbhardwaj: have modified the implementation

internal/lib/cron/cron.go
Mryashbhardwaj: yes

client/cmd/internal/logger/color.go
Mryashbhardwaj: no, i'll check the existing implementation

Mryashbhardwaj: removed the file

job/service.go
```diff
@@ -694,6 +739,25 @@ func (srv *Service) fetchSpecsForGivenJobNames(ctx context.Context, projectSpec
 	return jobSpecs, nil
 }
 
+func (srv *Service) IsJobDestinationDuplicate(ctx context.Context, jobSpec models.JobSpec) (string, error) {
```
arinda-arif, line 742: ideally, function name with `Is` prefix indicates the function doing the check and returning a boolean value. let's rename this to something else, for example `GetJobNamesWithDuplicateDestination` or feel free to suggest if you have better name.

we also don't need to pass the whole JobSpec, `fullName` and `destination` should be enough.

internal/lib/cron/cron.go
Mryashbhardwaj: removed

arinda-arif (approved):
LGTM. If possible you can add more test cases before merging.

Mryashbhardwaj:
> LGTM. If possible you can add more test cases before merging.

how about adding more tests after the final refactor

PR #553, opened 2022-09-01
fix: delete cascade job_run and metric info on job hard deletion 

### LINKED ISSUE #551: Bug: Unable to recreate deleted jobs with same name
**Describe the bug**
Unable to hard delete from `job` table due to foreign key constraint `job_run_job_id_fkey` on table `job_run`

**To Reproduce**
Steps to reproduce the behaviour:
1. Delete a job, (this will result in soft delete)
2. Recreate a new job with the same name (to be able to create a new job, we now have to hard delete old (soft deleted) jobs with same name if any)
3. Job recreate will fail

**Expected behaviour**
One should be able to recreate jobs with same names

Mryashbhardwaj:
fixes #551

### DIFF
--- a/server/dependencies.go
+++ b/server/dependencies.go
@@ -47,8 +47,9 @@ func (fac *replayWorkerFact) New() job.ReplayWorker {
 
 // namespaceJobSpecRepoFactory stores raw specifications
 type namespaceJobSpecRepoFactory struct {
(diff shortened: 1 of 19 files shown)


sravankorumilli:
Delete cascade feature should be used instead

PR #555, opened 2022-09-02
refactor: cleanup minor part 3

deryrahman:


job/service.go
```diff
@@ -441,13 +439,17 @@ func (srv *Service) GetDependencyResolvedSpecs(ctx context.Context, proj models.
 }
 
 // do other jobs depend on this jobSpec
-func (srv *Service) isDependency(ctx context.Context, jobSpec models.JobSpec) (bool, error) {
+func (srv *Service) getDependentJobNames(ctx context.Context, jobSpec models.JobSpec) ([]string, error) {
 	// inferred and static dependents
 	dependentJobs, err := srv.jobSpecRepository.GetDependentJobs(ctx, &jobSpec)
 	if err != nil {
-		return false, fmt.Errorf("unable to check dependents of job %s", jobSpec.Name)
+		return []string{}, fmt.Errorf("unable to check dependents of job %s", jobSpec.Name)
```
irainia, line 446: better return nil to avoid unnecessary allocation

job/service.go
```diff
@@ -441,13 +439,17 @@ func (srv *Service) GetDependencyResolvedSpecs(ctx context.Context, proj models.
...
 	}
-	return len(dependentJobs) > 0, nil
+	jobNames := []string{}
+	for _, job := range dependentJobs {
+		jobNames = append(jobNames, job.Name)
```
irainia, line 450: rather than using `append`, maybe we can modify it into:
```go
jobNames[i] = job.Name
```
this is to avoid capacity unnecessarily updated

irainia (changes requested):

irainia (approved):

PR #562, opened 2022-09-06
feat: improve logs on cyclic deps

deryrahman:


core/tree/multi_root_tree.go
```diff
@@ -61,36 +65,74 @@ func (t *MultiRootTree) IsCyclic() error {
...
 			}
 			if cyclicErr != nil {
 				return cyclicErr
 			}
 
 			_, childAlreadyInPath := pathMap[child.GetName()] // 1 -> 2 -> 1
 			if childAlreadyInPath && pathMap[child.GetName()] {
-				cyclicErr = fmt.Errorf("%w: %s", ErrCyclicDependencyEncountered, root.GetName())
+				*paths = append(*paths, child.GetName())
+				cyclicErr = fmt.Errorf("%w: %s", ErrCyclicDependencyEncountered, stringifyPaths(*paths))
```
arinda-arif, line 105: not sure if `stringify` term is correct for this as it is already an array of strings. should we call it `joinPaths` or you have another suggestion?

core/tree/multi_root_tree.go
```diff
@@ -61,36 +65,74 @@ func (t *MultiRootTree) IsCyclic() error {
...
+
+func (t *MultiRootTree) isCyclic(root *TreeNode, visited map[string]bool) error {
+	pathMap := make(map[string]bool)
+	paths := []string{}
+	return t.hasCycle(root, visited, pathMap, &paths)
+}
+
 // runs a DFS on a given tree using visitor pattern
-func (t *MultiRootTree) hasCycle(root *TreeNode, visited, pathMap map[string]bool) error {
+func (t *MultiRootTree) hasCycle(root *TreeNode, visited, pathMap map[string]bool, paths *[]string) error {
```
arinda-arif, line 85: tbh it is quite confusing to have `visited`, `pathMap`, and `paths`. however, we can just keep the scope of this as improving logs (since this is quite important) and not refactor anything related to the logic. we can create another issue for it and revisit this later. wdyt @deryrahman @irainia ?

core/tree/multi_root_tree.go
deryrahman: `printPaths`? or maybe `prettifyPaths`? As it's printing the array paths in tree fashion(?)

core/tree/multi_root_tree.go
irainia: IMHO, we can implement function within function to achieve recursion as required. With this approach, the caller of the main function, perhaps in this case `hasCycle`, does not need to think about how to initiate `pathMap` or any other arguments. Ideally, `hasCycle` should only accept one argument, which is `root`, and return only one result, which is `error`. Though, again, we may need to reconsider whether to use `hasCycle` as the function name. Because, when I read `hasCycle`, I would assume the return value is boolean. Similar to `isValid`, which is boolean and `Validate` which returns error. Maybe we can use `validateCyclic` or any other name that express more intent.

core/tree/multi_root_tree.go
```diff
@@ -61,36 +65,74 @@ func (t *MultiRootTree) IsCyclic() error {
...
+func (t *MultiRootTree) getSortedKeys() []string {
+	keys := []string{}
+	for k := range t.dataMap {
+		keys = append(keys, k)
+	}
+	sort.Strings(keys)
+	return keys
+}
+
+func (t *MultiRootTree) isCyclic(root *TreeNode, visited map[string]bool) error {
```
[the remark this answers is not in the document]
irainia: IMHO, we can implement function within function to achieve recursion as required. With this approach, the caller of the main function, perhaps in this case hasCycle, does not need to think about how to initiate pathMap or any other arguments. Ideally, hasCycle should only accept one argument, which is root, and return only one result, which is error. Though, again, we may need to reconsider whether to use hasCycle as the function name. Because, when I read hasCycle, I would assume the return value is boolean. Similar to isValid, which is boolean and Validate which returns error. Maybe we can use validateCyclic or any other name that express more intent.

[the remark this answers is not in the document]
irainia: IMHO, we can implement function within function to achieve recursion as required. With this approach, the caller of the main function, perhaps in this case `hasCycle`, does not need to think about how to initiate `pathMap` or any other arguments. Ideally, `hasCycle` should only accept one argument, which is `root`, and return only one result, which is `error`. Though, again, we may need to reconsider whether to use `hasCycle` as the function name. Because, when I read `hasCycle`, I would assume the return value is boolean. Similar to `isValid`, which is boolean and `Validate` which returns error. Maybe we can use `validateCyclic` or any other name that express more intent.

arinda-arif (approved):

Issue #335, opened 2022-05-11
Refactor Logger


irainia:
Differently formatting message, such as coloring message, could help the user to notice on something important. But, the current implementation of providing log could be improved. As observed, the following is example of the current implementation of using log:

```go
l.Info(coloredNotice("some message")) // with l is for log.Logger
```
_the above example is taken from [here](https://github.com/odpf/optimus/blob/main/cmd/deploy.go#L60)_

There's also another example like the following:

```go
l.Warn(coloredNotice(fmt.Sprintf("some message %s", actualMessage))) // with l is for log.Logger
```
_the above example is taken from [here](https://github.com/odpf/optimus/blob/main/cmd/admin_build_instance.go#L143)_

The above example uses functions, as defined:

```go
...
coloredNotice  = fmt.Sprintf
coloredError   = fmt.Sprintf
coloredSuccess = fmt.Sprintf
...
```
_taken from [here](https://github.com/odpf/optimus/blob/main/cmd/commands.go#L17)_

With the current approach:
* we need to handle formatting in separate fashion, like coloring in one function and formatting string message in another
* although there might be a justifying reason, inconsistency in formatting (such as color) could rise (e.g. Info could be white or yellow or any other color)
* in certain cases, we need to access the custom formatting function (like coloring function) if we want to use it outside of the `cmd` package (as currently majority of it is used here)

Our proposal is to address the above concerns. A more detailed approach could be finalized, but on the high level, **log should handle any formatting (apart from string format value) behind the scene** (not limited to coloring for standard output). For example in term of coloring:

```go
logger.Info("some message from: %s", userName) // the message is written as plain white wherever it's called
logger.Warn("request message is halted by: %s", adminName) // the message is written as plain yellow wherever it's called
logger.Error("message expired as: %s", expirationReason) // the message is written as plain red wherever it's called
```

ravisuhag:
@irainia salt has a printer module which provides these func. I would recommend us to use salt package for all cli outputs.

irainia:
> @irainia salt has a printer module which provides these func. I would recommend us to use salt package for all cli outputs.

Sure, we can use any required libraries, including salt.

PR #565, opened 2022-09-07
fix: unproper logger print

deryrahman:


cmd/deploy/deploy.go
```diff
@@ -86,7 +86,7 @@ func (d *deployCommand) PreRunE(_ *cobra.Command, _ []string) error {
 }
 
 func (d *deployCommand) RunE(_ *cobra.Command, _ []string) error {
-	d.logger.Info("Registering project [%s] to [%s]", d.clientConfig.Project.Name, d.clientConfig.Host)
+	d.logger.Info(fmt.Sprintf("Registering project [%s] to [%s]", d.clientConfig.Project.Name, d.clientConfig.Host))
```
irainia, line 89: Let's not use `fmt.Sprintf` here since doing that defeats the purpose of the previous issue for this update, mentioned [here](https://github.com/odpf/optimus/issues/335). The issue's goal is to remove this `fmt.Sprintf` or the like entirely. Meaning, we aim it to be:
```go
d.logger.Info("Registering project [%s] to [%s]", d.clientConfig.Project.Name, d.clientConfig.Host)
```
not
```go
d.logger.Info(fmt.Sprintf("Registering project [%s] to [%s]", d.clientConfig.Project.Name, d.clientConfig.Host))
```

irainia (changes requested):

cmd/deploy/deploy.go
irainia: updated

irainia (approved):

PR #606, opened 2022-09-19
fix: bug on resource that is not deployed

deryrahman:
Prototyping of atomicity property on create / update resource


sravankorumilli:
Closing this as it is not needed will be taken care as part of resource bounded context.

Issue #570, opened 2022-09-08
Move Yaml specs and their reader to client side


sbchaos:
**Description**
Currently the code for cli to read the yaml configuration related to job and resource are present along with server side code. We need to move these to client side, to establish the clear contract for input by users.

**Acceptance Criteria**

The client side code which defines user contracts as yaml should be moved to cli side of the code and not be used by the server side.

**Out of Scope**

1. Changing of any api contracts

**Tech Details**
Post analysis Capture the details of the work needed to finish the task.

- [ ] Simplify reading of different spec files
- [ ] Move reading of yaml spec to client, remove usage of them from server, eg local.Job
- [ ] Break away from common interface for Postgres and local store, move local store to client side
- [ ] Remove filesystem cache and implement file/folder reading
- [ ] Do not use server side models on client, convert yaml spec to proto contracts directly

irainia:
Result of discussion about the approach on 2022-09-21:

* we expecte proto not to be changed
* split the bigquery specific structure, such as `BQTableMetadata` into smaller one
* validation happens in server, meaning client is not expected to do such
* client only read, convert to proto directly, and send to the server
* the implementation addresses both job and resource

PR #616, opened 2022-09-23
refactor: re-work local spec implementation

irainia:
This PR is to address [this issue](https://github.com/odpf/optimus/issues/570). This PR is still in progress. So, further implementation will be applied later along with the detailed description of this PR.

Tracker:
- [x] read write job spec from and to yaml file + assets
- [x] read write resource spec from and to yaml file
- [x] convert job spec to proto
- [x] convert resource spec to proto
- [x] integrating with cmd
- [x] clean up unused code


deryrahman:
NB: we removed the cli render, since to achieve render process, we need to create render api on server side, which is out of this PR's scope. The capability of render process will be taken care on other PR

client/local/specio/resource_spec_io.go
```diff
@@ -0,0 +1,76 @@
...
+		specs[i] = spec
+	}
+	return specs, nil
+}
+
+func (r resourceSpecReadWriter) ReadByName(rootDirPath, name string) (*model.ResourceSpec, error) {
+	if name == "" {
+		return nil, errors.New("name is empty")
+	}
+	allSpecs, err := r.ReadAll(rootDirPath)
```
sbchaos, line 56: Are we reading all the specs here to read only a single spec. Is there any other reliable way to find the specs ?

client/local/specio/resource_spec_io.go
```diff
@@ -0,0 +1,76 @@
...
+	if rootDirPath == "" {
+		return nil, errors.New("root dir path is empty")
+	}
+
+	specDirPaths, err := internal.DiscoverSpecDirPaths(r.specFS, rootDirPath, r.referenceSpecFileName)
+	if err != nil {
+		return nil, fmt.Errorf("error discovering spec paths under [%s]: %w", rootDirPath, err)
+	}
+
+	specs := make([]*model.ResourceSpec, len(specDirPaths))
```
sbchaos, line 40: 👍

client/local/model/resource_spec.go
```diff
@@ -0,0 +1,32 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("error constructing spec pb: %w", err)
+	}
+	return &pb.ResourceSpecification{
+		Version: int32(r.Version),
+		Name:    r.Name,
+		Type:    r.Type,
+		Labels:  r.Labels,
+		Spec:    specPb,
+		Assets:  nil, // TODO: check if we really need assets
```
sbchaos, line 30: We can remove assets.

client/local/specio/resource_spec_io.go
irainia: agreed that reading all may be not the optimal solution we are looking for. but, the tricky part is, we are searching by the name. and no other place we can get the name information other than reading the whole file. the closest we can do is we can limit it. for example, we read the specs one by one and stop after finding what we are looking for. it can reduce the number of file we have to read. what do you think?

sbchaos (approved):

PR #621, opened 2022-09-26
fix : handle deletions for both namespace id vs namespace name folders

Mryashbhardwaj:


ext/scheduler/airflow2/airflow.go
```diff
@@ -233,7 +230,7 @@ func (s *scheduler) ListJobs(ctx context.Context, namespace models.NamespaceSpec
 		return jobs, nil
```
Mryashbhardwaj, line 230: `opts.OnlyName` is always true in the codebase, should we remove this check? and the following code

job/deployer.go
```diff
@@ -113,27 +114,39 @@ func (d *deployer) completeJobDeployment(ctx context.Context, jobDeployment mode
...
+			sourceJobNames = append(sourceJobNames, jobSpec.Name)
+		}
+		jobsToDelete := setSubtract(destJobNames, sourceJobNames)
+		if len(jobsToDelete) > 0 {
+			if err := d.batchScheduler.DeleteJobs(ctx, nsDirectoryIdentifier, namespaceSpec, jobsToDelete, nil); err != nil {
+				return err
+			}
+		}
+
+		err = d.batchScheduler.DeleteDagsDirectoryIfEmpty(ctx, nsDirectoryIdentifier, namespaceSpec)
```
Mryashbhardwaj, line 144: only do this for the namespaceID folder

sravankorumilli (approved):

PR #629, opened 2022-10-07
fix: keep things backward compatible for other deployments depending on the upgraded optimus server.

sravankorumilli:


api/handler/v1beta1/job_run.go
```diff
@@ -278,11 +278,14 @@ func (sv *JobRunServiceServer) JobRun(ctx context.Context, req *pb.JobRunRequest
 }
 
 func (*JobRunServiceServer) GetWindow(_ context.Context, req *pb.GetWindowRequest) (*pb.GetWindowResponse, error) {
+	version := 1
 	if err := req.GetScheduledAt().CheckValid(); err != nil {
 		return nil, status.Errorf(codes.Internal, "%s: failed to parse schedule time %s", err.Error(), req.GetScheduledAt())
 	}
-
-	window, err := models.NewWindow(int(req.Version), req.GetTruncateTo(), req.GetOffset(), req.GetSize())
+	if req.Version != 0 {
```
irainia, line 285: It might be not the best practice, but what do you think if we put the default window version assignment inside `models.NewWindow` function? So, it would be like the following:

```go
func NewWindow(version int, truncateTo, offset, size string) (Window, error) {
	if version <= 1 {
		return windowV1{truncateTo: truncateTo, offset: offset, size: size}, nil
	}
        ...
}
```

_notice the checking `if version <= 1`_

This is because window initialization usage is being used in three other places outside this `GetWindow` method:
* api adapter [FromJobProto](https://github.com/odpf/optimus/blob/release-v0.4/api/handler/v1beta1/adapter.go#L53)
* local adapter [ToSpec](https://github.com/odpf/optimus/blob/release-v0.4/store/local/job_spec_adapter.go#L404)
* postgres adapter [ToSpec](https://github.com/odpf/optimus/blob/release-v0.4/store/postgres/adapter.go#L235)

The pros, we can be sure that the object of window will have version 1 by default if not specified. The cons, though, it does not enforce the caller to explicitly define the window version. This cons needs consideration whether such action is considered safe or not.

sravankorumilli: as the fix is to provide a backward compatibility only to this api & which needs to be removed in future releases lets avoid keeping this a norm in all places.

PR #631, opened 2022-10-07
fix: show error message regardless verbosity condition

deryrahman:


client/cmd/internal/logger/printer.go
```diff
@@ -6,15 +6,16 @@ import (
 	pb "github.com/odpf/optimus/protos/odpf/optimus/core/v1beta1"
 )
 
-func PrintLogStatus(logger log.Logger, logStatus *pb.Log) {
-	switch logStatus.GetLevel() {
-	case pb.Level_LEVEL_INFO:
+func PrintLogStatus(logger log.Logger, logStatus *pb.Log, verbose bool) {
```
irainia, line 9: Let's avoid using flag arguments in the function as they are one of smell codes. Instead of sending verbosity status into this function, what do you think if we split the function itself into two? One of them is for printing with verbose (for example `PrintLogStatusVerbose`) and another without verbose (such as `PrintLogStatus`). Think of it like `fmt.Print` and `fmt.Println`.

irainia (changes requested):
We can have a separate discussion if necessary.

client/cmd/internal/logger/printer.go
deryrahman: Got it, so control flow will be handled in the outside PrintLog func

PR #643, opened 2022-10-19
refactor: mark namespace name mandatory in tenant

sbchaos:


core/tenant/handler/v1beta1/secret.go
```diff
@@ -14,10 +14,10 @@ import (
 )
 
 type SecretService interface {
-	Save(context.Context, tenant.Tenant, *tenant.PlainTextSecret) error
```
sravankorumilli, line 17: except in GetSecretsInfo we could have used tenant everywhere else right? Is your reason for that is to maintain the uniformity across all functions?

sbchaos: All api have namespace as optional.

`GetSecretsInfo` only needs project, others change behaviour depending on whether namespace is present or not. We have different queries for both condition.

Issue #622, opened 2022-09-27
fix log formatting with salt logger 


Mryashbhardwaj:
**Describe the bug**
current logger does not support argumentative logging as used on https://github.com/odpf/optimus/blob/main/job/deployer.go#L51 
we need to format this log using `fmt.Sprintf` 

**To Reproduce**
Steps to reproduce the behaviour:
1. trigger deployment that is about to get failed 
2. check the log msg, it does not capture the error and deployment details

**Expected behaviour**
the logs should be formatted

camclark:
Hey team, interesting issue. 
Just stumbled on your project; mind if I have this issue assigned to me?

camclark:
Been playing around with the repo and I think I'm missing something to get local working.

tldr: Followed local guide, but unable to get ./optimus serve to work from default configs. Under the impression localhost should work near out of the box.

Followed steps as per: https://github.com/odpf/optimus#running-locally
```
$ ./optimus version
Client: v0.4.0-rc.6-next-17053a1

$ ./optimus serve  
WARN[0000] Shutting down server                         
INFO[0000] Server shutdown complete                     
Error: unable to create server: serve.ingress_host: required config missing
🔥 unable to complete request successfully
``` 

Since server isn't starting follow next steps within: https://odpf.github.io/optimus/docs/getting-started/configuration/

```
$ ./optimus init   
```

From here generated the optimus.yaml using values from this configuration page
```
version: 1
log:
  level: INFO
  format: ""
host: localhost:9100
project:
  name: local-project
  config: {}
namespaces:
- name: namespace1
  config: {}
  job:
    path: namespace1/jobs
  datastore:
  - type: bigquery
    path: namespace1/resources
    backup: {}
```

attempt deploy:
```
$ ./optimus deploy
Initializing client plugins
initialization finished!

Registering project [local-project] to [localhost:9100]
Error: Unable to reach optimus server at localhost:9100, this can happen due to following reasons:
1. Check if you are connected to internet
2. Is the host correctly configured in optimus config
3. Is Optimus server currently unreachable
🔥 unable to complete request successfully
```

See server issue and scroll down on that page to: server config

copy config.sample.yaml and uncomment as config.yaml

```
version: 1

# logging configuration
log:
  # debug, info, warning, error, fatal - default 'info'
  level: info

########################################
# SERVER CONFIG
########################################

# for configuring optimus service
serve:
 # port to listen on
 port: 9100

 # host to listen on
 host: localhost

[...]

Issue #574, opened 2022-09-08
Create resource management bounded context 


sbchaos:
**Description**
Create a module to handle and encapsulate the logic related to resource management.

**Acceptance Criteria**

- [x] All handlers, service for managing resources are present in this bounded context.
- [x] Use new resource models defined in the bounded context for the BQ interactions.
- [x] Client should still rely on the same apis & there shouldn't be any change in the api contracts.
- [ ] Capture the relevant metrics for all apis & update the dashboards accordingly.
- [ ] The api should provide enough information for users to understand the issue & the logs should be meaningful enough to debug any issue happening on the server side.

**Scope**

- [x] Introduction of the new bounded context with all the handlers & services for resource handling.
- [x] Introduction of new repositories to work with the new models being introduced as part of this BC.
- [ ] Improve the debuggability through logging, capturing the right metrics & updating the dashboards.

**Out of Scope**

- Cleanup 

Depends on : [573](https://github.com/odpf/optimus/issues/573)

irainia:
## Scenario

### Create Resource

* if resource does not exist in database, check store:
    * if exist, then we consider it as externally managed and we reject the request
    * if does not exist, then we allow for creation to both database and to store
* if resource exist in database, check its status related to store:
    * if the status says it exists, then we reject the request
    * if the status says it does not exist, then check store:
        * if exist in store, then update status resource related to store in database accordingly
        * if does not exist in store, then we replace the one in database and create to store

### Update Resource

* if resource does not exist in database, then reject the request
* if resource exist in database, then check its status related to store:
    * if status says it exists, then update both the database and the store
    * if status says it does not exist, then check store:
        * if exist in store, then update status related to store accordingly
        * if does not exist in store, then we reject the request

### Deploy

* deployment is done to all the incoming resources
* no deletion is required during deployment
* we don't skip any resource since we aim to make it incremental in the later implementations
* the operation consists of create process or update process, depending on whether the related resources are in database and/or store or not
* partial resource deployment is expected

Issue #533, opened 2022-08-22
deploy resource ignores the resources once its failed to create or update


SumitAgrawal03071989:
**Describe the bug**
Currently, while deploying the resources, e.g. big query table, if any resource failed to deploy because of any reason, then in next run the resource gets ignored.

**To Reproduce**
1. create a big query table as a resource in dwh repository
2. do not register or delete the secrets TASK_BQ2BQ and DATASTORE_BIGQUERY and STORAGE secret 
3. run optimus deploy to deploy the resources 
    - here the deploy command will fail, with error that secrets are not configured.
5. register TASK_BQ2BQ and DATASTORE_BIGQUERY and STORAGE with appropriate service account.
6. run the deploy command again.
7. the table will be ignored with message `{table_name} is skipped because incoming resource is the same as existing
resources`

**Expected behavior**
Table which were unable to sync to big query should be part of optimus deploy

PR #650, opened 2022-10-31
refactor: resource bounded context

irainia:
This PR is to address [this issue](https://github.com/odpf/optimus/issues/574). Changes on this PR:

* re-implement resource service but with the similar behaviour as the previous version
* re-build resource table along with its repository implementation but still retain the old one
* add status on resource to handle bug resource not being updated, which at the same time address [this issue](https://github.com/odpf/optimus/issues/533)
* re-implement bigquery datastore for easier readability
* add multi-error, which can be used to handle multiple errors
* add a wrapper for server logger, which also address [this issue](https://github.com/odpf/optimus/issues/622)
* add additional metric for batch update (or resource deployment)

Note:
* old implementations such as service and repository are still maintained in this PR, but will be deprecated in another PR


core/resource/resource.go
```diff
@@ -0,0 +1,309 @@
...
+		ds.Dataset = r.dataset
+		return ds.Validate()
+
+	default:
+		return errors.InvalidArgument(EntityResource, "unknown kind")
+	}
+}
+
+func (r *Resource) Equal(incoming *Resource) bool {
+	if r == nil || incoming == nil {
```
sbchaos, line 187: `r` cannot be nil, as it is a method call, we will have nil pointer error. So we can omit that condition.

irainia: Technically, `r` can be nil. If `r` is nil, we will encounter panic if we remove this condition. Now, if we assume that `r` cannot be nil because of how we designed Optimus, then we may as well assume that `incoming` cannot be nil. If that's the case, then we can just remove `nil` check condition, such as in line `187` mentioned here.

core/resource/handler/v1beta1/resource.go
```diff
@@ -0,0 +1,341 @@
...
+		rh.l.Error("Error while deploying namespaces: [%s]", namespacesWithError)
+		return fmt.Errorf("error when deploying: [%s]", namespacesWithError)
+	}
+	return nil
+}
+
+func (rh ResourceHandler) ListResourceSpecification(ctx context.Context, req *pb.ListResourceSpecificationRequest) (*pb.ListResourceSpecificationResponse, error) {
+	store, err := resource.FromStringToStore(req.GetDatastoreName())
+	if err != nil {
+		return nil, errors.GRPCErr(errors.InvalidArgument(resource.EntityResource, "invalid datastore name"), "invalid list resource request")
```
sbchaos, line 143: `err` should be InvalidArgument error, we should use the same.

core/resource/handler/v1beta1/resource.go
```diff
@@ -0,0 +1,341 @@
...
+	}
+	var me *errors.MultiError
+	if errors.As(err, &me) {
+		for _, e := range me.Errors {
+			writeError(logWriter, e)
+		}
+	} else {
+		var de *errors.DomainError
+		if errors.As(err, &de) {
+			logWriter.Write(writer.LogLevelError, de.DebugString())
```
sbchaos, line 258: Debug string should be printed only on the server side. we should not send this to the client side. It will contain errors from external system.

irainia: I agree that external error messages will be included here. However, if we don't send those messages, the user will get so minimal error message which I'm sure they will get confused on what is wrong with their request. Especially when they do deployment, if there are some migrations that involve many resources to be created or updated, it will be tricky for them to understand the encountered errors.

core/resource/resource.go
```diff
@@ -0,0 +1,255 @@
...
+func FromExisting(existing *Resource, opts ...FromExistingOpt) *Resource {
+	output := &Resource{
+		name:         existing.name,
+		kind:         existing.kind,
+		dataset:      existing.dataset,
+		tenant:       existing.tenant,
+		spec:         existing.spec,
+		metadata:     existing.metadata,
+		status:       existing.status,
+		existInStore: existing.existInStore,
```
sbchaos, line 239: Why do we need this condition ?

irainia: during the request operation, at some points we will need to confirm whether the particular resource actually exist in datastore or not. take an example when we want to create a table in bigquery. during the create request, we will need to do this check on the bigquery, and accept or reject the request based on it. we need to do so because sometime the resource could already exist, be it from manual creation or "other" optimus or whatnot. this `existInStore` is to help reducing the request to do the check in the actual bigquery.

sbchaos (approved):

core/resource/resource.go
```diff
@@ -233,7 +233,7 @@ func (r *Resource) MarkSkipped() error {
 }
 
 func (r *Resource) MarkToCreate() error {
-	if r.status == StatusValidationSuccess || r.status == StatusCreateFailure {
+	if r.status == StatusValidationSuccess {
```
irainia, line 236: What is the reason of changing the requirement to transition status into `to_create` only status `success`? I think it's a bit different than the last discussion.

core/resource/resource.go
```diff
@@ -243,7 +243,7 @@ func (r *Resource) MarkToCreate() error {
 
 func (r *Resource) MarkToUpdate() error {
 	switch r.status {
-	case StatusValidationSuccess, StatusUpdateFailure, StatusSuccess, StatusExistInStore:
+	case StatusValidationSuccess:
```
irainia: Also for `to_update`, perhaps we can discuss about this as it's a bit different than the last discussion.

core/resource/service/resource_service.go
```diff
@@ -146,51 +137,68 @@ func (rs ResourceService) Deploy(ctx context.Context, tnnt tenant.Tenant, store
...
+			continue
 		}
+
+		if resource.StatusForToCreate(existing.Status()) {
+			_ = incoming.MarkToCreate()
+		} else if resource.StatusForToUpdate(existing.Status()) {
+			_ = incoming.MarkToUpdate()
+		}
+
+		err := rs.repo.Update(ctx, incoming)
```
irainia, line 192: Maybe we can move this `Update` repo call into outside of this function `getResourcesToBatchUpdate`; or we change the name `getResourcesToBatchUpdate` into something else like `updateResourcesForDeploy`. Because if we don't do any changes, then the function will not reflect what it actually does.

irainia (commented):
Just let me know if a separate call is required.

core/resource/resource.go
sbchaos: The current implementation was wrong. Let's have a call if you have some confusion. Only the resources in validation success can be marked as to_create. It will not be done for existing resources.

core/resource/resource.go
sbchaos: Sure, let's have a discussion as the current implementation was wrong.

core/resource/service/resource_service.go
sbchaos: I disagree, let's discuss that.

PR #651, opened 2022-11-02
feat: improve dev setup

deryrahman:
Changes:
- [x] `make apply` using parameterize value of `DAGS_PATH`, and `OPTIMUS_SERVE_PORT`
- [x] provide dummy echo plugin as an example
- [x] add script to load secret defined in `setup.yaml`
- [x] script to install plugins through artifacts defined in `setup.yaml`


dev/Makefile
```diff
@@ -2,34 +2,37 @@ EXPECTED_CONTEXT=colima
 NAMESPACE=optimus-dev
 OPTIMUS_ODPF_CHART=odpf/optimus
 
+DAGS_PATH?=./dags
```
smarch-int, line 5: using abspath is decent but only `$HOME/*` and `/tmp/colima/*` will work. Other paths won't work.

dev/plugins/optimus-plugin-echo.yaml
```diff
@@ -0,0 +1,9 @@
+name: echo
+description: Echo sample plugin for optimus
+plugintype: task
+pluginversion: 0.1.0
```
smarch-int, line 4: image?

dev/Makefile
```diff
@@ -2,34 +2,37 @@ EXPECTED_CONTEXT=colima
...
+OPTIMUS_PLUGINS_PATH?=./plugins
```
smarch-int, line 6: Here the assumption is to plugins are copied to this folder (extra step)

dev/load_secret.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/sh
+
+PROJECT=${PROJECT:-project-a}
+SECRET_NAME=${SECRET_NAME:-BQ_SERVICE_ACCOUNT}
```
smarch-int, line 4: how to add multiple secrets ?

smarch-int (changes requested):

dev/Makefile
deryrahman: I made an assumption that user by default always create a folder inside `$HOME/*`, and any relative path will be expanded to absolute path on this loc https://github.com/odpf/optimus/blob/57725f11e1a216b7f8a3aa8baaff87a6c28946b4/dev/Makefile#L20

dev/load_secret.sh
deryrahman: added feat to load multiple secrets

dev/Makefile
deryrahman: remove this one, use artifacts list inside setup.yaml instead

smarch-int (changes requested):

dev/README.md
```diff
@@ -1,34 +1,54 @@
...
++ `mkdir project-a`
++ `cd project-a`
++ `optimus init`
++ `optimus project register`
++ `optimus plugin sync`
++ then load the secret. [ref](#load-secrets)
+
+### Mounting the plugins
++ define plugin artifacts on `setup.yaml` under section `plugins`
++ pass the variable OPTIMUS_PLUGIN_ARTIFACTS when [applying optimus](#setup) (comma separated delimiter). eg.
```
smarch-int, line 37: why to pass the var when we are listing the plugins in setup.yaml ?

deryrahman: updated, committed the new one

smarch-int (approved):
LGTM!

PR #682, opened 2022-12-07
Feature/scheduler upstream

Mryashbhardwaj:


client/cmd/job/job.go
```diff
@@ -31,7 +31,6 @@ func NewJobCommand() *cobra.Command {
 		NewRefreshCommand(),
 		NewRunListCommand(),
 		NewValidateCommand(),
-		NewJobRunInputCommand(),
```
sbchaos, line 34: Are we removing this command ?

Mryashbhardwaj: this is moved to scheduler BC

PR #684, opened 2022-12-08
fix: port commits from 0.4 to main

sravankorumilli:


sbchaos (approved):

ext/scheduler/airflow2/resources/base_dag.py
```diff
@@ -116,7 +116,7 @@
 )
 {{- end }}
 JOB_DIR = "/data"
-IMAGE_PULL_POLICY="Always"
+IMAGE_PULL_POLICY="IfNotPresent"
```
smarch-int, line 119: @sravankorumilli  any issue with `Always` ?

smarch-int: some executors like bgsensor have only latest tags

PR #699, opened 2023-01-04
refactor: move plugin models to sdk

sbchaos:


sravankorumilli (approved):

plugin/yaml/plugin.go
```diff
@@ -100,15 +100,15 @@ func NewPluginSpec(pluginPath string) (*PluginSpec, error) {
 		return nil, err
 	}
 	var plugin PluginSpec
-	if err := yaml.UnmarshalStrict(pluginBytes, &plugin); err != nil {
+	if err := yaml.Unmarshal(pluginBytes, &plugin); err != nil { // TODO: check if strict marshal is required
```
smarchint, line 103: why was this removed ? @sbchaos

sbchaos: New plugin model has few items removed, so made it less strict for migration.

PR #703, opened 2023-01-05
refactor: remove secret type

sbchaos:


internal/store/postgres/tenant/secret_repository.go
```diff
@@ -126,8 +112,8 @@ func (s SecretRepository) Save(ctx context.Context, tenantSecret *tenant.Secret)
 	}
 
 	insertSecret := `INSERT INTO secret (name, value, type, project_name, namespace_name, created_at, updated_at)
```
sravankorumilli, line 114: we have type here still

core/tenant/secret.go
```diff
@@ -8,10 +8,6 @@ const (
 	SecretStorageKey    = "STORAGE"
 	SecretSchedulerAuth = "SCHEDULER_AUTH"
 
-	// SystemDefinedSecret TODO: get rid of system defined secrets
-	SystemDefinedSecret SecretType = "system"
-	UserDefinedSecret   SecretType = "user"
-
 	SecretTypeSystemDefinedPrefix = "_OPTIMUS_"
```
sravankorumilli, line 11: I think we are not using this prefix also, we can remove this

core/tenant/secret.go
```diff
@@ -8,10 +8,6 @@ const (
 	SecretStorageKey    = "STORAGE"
 	SecretSchedulerAuth = "SCHEDULER_AUTH"
```
sravankorumilli, line 9: Can you add the alert secret to be a predefined secret as well, doesn't have much significance but a place where we need what all secrets are part of infra setup.

sravankorumilli (changes requested):

internal/store/postgres/tenant/secret_repository.go
sbchaos: Needs a migration to remove the column. Will add it.

sravankorumilli (approved):

PR #712, opened 2023-01-09
fix: handle deleted jobs in scheduler BC

Mryashbhardwaj:


ext/scheduler/airflow/airflow.go
```diff
@@ -87,8 +87,10 @@ func (s *Scheduler) DeployJobs(ctx context.Context, tenant tenant.Tenant, jobs [
 	}
 	defer bucket.Close()
 
-	bucket.WriteAll(spanCtx, filepath.Join(jobsDir, baseLibFileName), SharedLib, nil)
-
+	err = bucket.WriteAll(spanCtx, filepath.Join(jobsDir, baseLibFileName), SharedLib, nil)
+	if err != nil {
+		return err
```
sbchaos, line 92: Wrap in meaning msg from optimus, It will only return the library error.

Mryashbhardwaj: Done

Issue #715, opened 2023-01-10
Static and inferred downstream job creates duplicate sensor


sbchaos:
**Describe the bug**
When a job is specified as a dependency in a job, as well as the job is inferred by optimus, we end up having 2 different sensors for the same job.

**Expected behavior**
Only one sensor should be created if the job is static or inferred.

PR #719, opened 2023-01-11
fix: duplicated upstream issue & job BC unclear logs

arinda-arif:
- fixing https://github.com/odpf/optimus/issues/715
- remove inaccurate logs saying upstream resolved externally while in fact there is no external optimus registered
- improve logs in external upstream resolver by adding namespace prefix
- improve logs when static upstream is specified but empty
- improve logs when unable to delete jobs while doing replace all


internal/store/postgres/job/job_repository.go
```diff
@@ -359,7 +359,7 @@ func (JobRepository) toUpstreams(storeUpstreams []JobWithUpstream) ([]*job.Upstr
 	if err := me.ToErr(); err != nil {
 		return nil, err
 	}
-	return upstreams, nil
+	return job.Upstreams(upstreams).Deduplicate(), nil
```
sravankorumilli, line 362: check whether this deduplication is needed? & also the job inspect deduplication is not happening

sravankorumilli (changes requested):
Added comments

internal/store/postgres/job/job_repository.go
arinda-arif: have removed the multiple deduplication and add deduplication for job inspect case

arinda-arif:
@sravankorumilli please check the recent changes:
- deduplicating from the root (for `bulk resolve` cases in the repository layer, for `single resolve` cases after merging static & inferred). the second deduplication is the same, after merging the unresolved with resolved ones.
- supports deduplicating unresolved upstreams

sravankorumilli (approved):

PR #721, opened 2023-01-12
fix: cyclic deps for inferred and static

deryrahman:


core/job/service/job_service.go
```diff
@@ -322,20 +327,38 @@ func (j JobService) Validate(ctx context.Context, jobTenant tenant.Tenant, jobSp
 	if err != nil {
 		return err
 	}
-	jobMap := make(map[job.Name]*job.Job)
-	destinationToJobsMap := make(map[job.ResourceURN][]*job.Job)
-	for _, jobEntity := range jobsInProject {
-		jobMap[jobEntity.Spec().Name()] = jobEntity
-		if _, ok := destinationToJobsMap[jobEntity.Destination()]; !ok {
-			destinationToJobsMap[jobEntity.Destination()] = []*job.Job{}
+	jobsInProjectWithUpstreams, err := j.upstreamResolver.BulkResolveInternal(ctx, jobTenant.ProjectName(), jobsInProject)
```
arinda-arif, line 330: how about we also use `GetUnresolved` for these jobs that came from db? 
CMIIW I believe we don't really need to resolve the jobs and thus we can avoid introduce `BulkResolveInternal`

core/job/service/job_service.go
```diff
@@ -523,47 +546,40 @@ func (j JobService) generateJob(ctx context.Context, tenantWithDetails *tenant.W
...
 	dagTree := tree.NewMultiRootTree()
 	dagTree.AddNode(tree.NewTreeNode(rootJob))
 
 	for _, childJob := range jobMap {
 		childNode := findOrCreateDAGNode(dagTree, childJob)
-		for _, source := range childJob.Sources() {
-			if _, ok := destinationToJobMap[source]; !ok {
-				// source maybe from external optimus or outside project,
+		for _, upstream := range childJob.Upstreams() {
+			if _, ok := destinationToJobMap[upstream.Resource()]; !ok {
```
arinda-arif, line 563: Will this support jobs with static upstream that does not have destination/resource? as not all jobs have resource, for example jobs with python task

core/job/service/job_service_test.go
```diff
@@ -1874,21 +1923,57 @@ func TestJobService(t *testing.T) {
...
-			jobA := job.NewJob(sampleTenant, specA, "table-A", []job.ResourceURN{"table-C"})
+			jobA := job.NewJob(sampleTenant, specA, "table-A", []job.ResourceURN{"table-Z"})
 			jobB := job.NewJob(sampleTenant, specB, "table-B", []job.ResourceURN{"table-A"})
 			jobC := job.NewJob(sampleTenant, specC, "table-C", []job.ResourceURN{"table-B"})
+
+			jobsWithUpstreams := []*job.WithUpstream{
+				job.NewWithUpstream(
+					jobB,
+					[]*job.Upstream{
+						job.NewUpstreamUnresolvedInferred(job.ResourceURN("table-A")),
```
arinda-arif, line 1934: let's add or combine with static upstream to also test the cyclic due static upstream

arinda-arif (changes requested):

core/job/service/job_service.go
deryrahman: agree, just examine the code and yeah, use unresolved should be enough

core/job/service/job_service.go
deryrahman: Good point, in a current implementation, we use the resource as an identifier. In that case, if the job doesn't have a destination resource, then it won't be included in a cyclic deps detection.
Just thinking to use job name as identifier instead

arinda-arif: we might need to have both map, as for the unresolved inferred ones, the upstream won't have the job name yet.

arinda-arif: on second thought, I think having unresolved upstream might resulted in having a false negative case, where the cyclic is actually not there (as the upstream sensor will not be created if it is unresolved) for some cases. we can discuss more on the approach.

core/job/service/job_service.go
deryrahman: move Job with unresolved upstream to the job method

core/job/job.go
```diff
@@ -34,14 +34,58 @@ func (j Job) Spec() *Spec {
...
-func (j Job) GetName() string { // to support multiroot DataTree
+func (j Job) GetName() string {
 	return j.spec.name.String()
 }
 
 func (j Job) FullName() string {
 	return j.ProjectName().String() + "/" + j.spec.name.String()
 }
 
+func (j Job) GetJobWithUnresolvedUpstream() (*WithUpstream, error) {
```
arinda-arif, line 45: let's add test cases for `GetJobWithUnresolvedUpstream` and `GetJobsWithUnresolvedUpstreams`

core/job/service/job_service_test.go
```diff
@@ -2652,6 +2780,52 @@ type UpstreamResolver struct {
 	mock.Mock
 }
 
+// GetUnresolved provides a mock function with given fields: jobs
+func (_m *UpstreamResolver) GetUnresolved(jobs []*job.Job) ([]*job.WithUpstream, error) {
```
arinda-arif, line 2784: we can these `GetUnresolved` and `BulkResolveInternal` mocks

core/job/service/job_service.go
```diff
@@ -331,20 +334,42 @@ func (j JobService) Validate(ctx context.Context, jobTenant tenant.Tenant, jobSp
 	if err != nil {
 		return err
 	}
-	jobMap := make(map[job.Name]*job.Job)
-	destinationToJobsMap := make(map[job.ResourceURN][]*job.Job)
-	for _, jobEntity := range jobsInProject {
-		jobMap[jobEntity.Spec().Name()] = jobEntity
-		if _, ok := destinationToJobsMap[jobEntity.Destination()]; !ok {
-			destinationToJobsMap[jobEntity.Destination()] = []*job.Job{}
+	jobsInProjectWithUpstreams, err := job.Jobs(jobsInProject).GetJobsWithUnresolvedUpstreams()
```
arinda-arif, line 337: if possible we can extract out:
- fetching `jobsInProject`
- getting `jobsInProjectWithUpstreams`
- maps preparation
into a single function to make `Validate` easier to read.

arinda-arif (approved):
added only minor feedback. feel free to merge post that. thanks!

PR #722, opened 2023-01-13
Fix: scheduler BC testing issues 

Mryashbhardwaj:


client/cmd/scheduler/run_input.go
```diff
@@ -56,7 +56,7 @@ func NewJobRunInputCommand() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:     "run-input",
 		Short:   "Fetch jobRunInput assets for a scheduled execution",
-		Example: "optimus job run-input <job_name> --output-dir </tmp> --scheduled-at <2021-01-14T02:00:00+00:00> --type <task|hook> --name <bq2bq> --project \"project-id\" --namespace \"namespace\" ",
+		Example: "optimus scheduler run-input <job_name> --output-dir </tmp> --scheduled-at <2021-01-14T02:00:00+00:00> --type <task|hook> --name <bq2bq> --project \"project-id\" --namespace \"namespace\" ",
```
sravankorumilli, line 59: why job to scheduler?

Mryashbhardwaj: because this command has been moved to scheduler bounded context

sravankorumilli: but from user point of view, user is dealing with jobs so it should be under job.
job is the right grouping imo

Mryashbhardwaj: this can be moved, from scheduler to job too, apart from the API one on one mapping, i do not have any other strong reasoning.

PR #723, opened 2023-01-13
refactor: use connection multiplexer for http and grpc server

sbchaos:


core/resource/service/resource_service.go
```diff
@@ -157,18 +157,18 @@ func (rs ResourceService) Deploy(ctx context.Context, tnnt tenant.Tenant, store
 			r.MarkValidationFailure()
 			continue
 		}
-		r.MarkValidationSuccess()
 
 		urn, err := rs.mgr.GetURN(r)
```
sravankorumilli, line 161: provide the error information to the client as well

protos/odpf/optimus/core/v1beta1/job_spec.swagger.json
```diff
@@ -88,6 +88,166 @@
             "in": "path",
             "required": true,
             "type": "string"
+          },
+          {
+            "name": "job.version",
```
sravankorumilli, line 93: all these changes are new which ideally shouldn't change with buf version

server/optimus.go
```diff
@@ -190,11 +204,28 @@ func (s *OptimusServer) setupHTTPProxy() error {
...
+		grpcLn := s.mux.MatchWithWriters(
+			cmux.HTTP2MatchHeaderFieldSendSettings("content-type", "application/grpc"))
+		if err := s.grpcServer.Serve(grpcLn); err != nil {
+			s.logger.Info("shutting down grpc server")
+		}
+	}()
+
+	go func() {
+		// Start listening on http server
+		httpL := s.mux.Match(cmux.HTTP1Fast())
```
sravankorumilli, line 217: check if it works for http2

sravankorumilli (changes requested):

protos/odpf/optimus/core/v1beta1/job_spec.swagger.json
sbchaos: Should not be part of changes. Reverting changes for buf update, maybe some issue.

core/resource/service/resource_service.go
sbchaos: Done

server/optimus.go
sbchaos: Used check for grpc and send else to http like before.

sravankorumilli (approved):

PR #724, opened 2023-01-15
fix: add missing template functions

sbchaos:


sravankorumilli:
@sbchaos can you add tests for these functions

sravankorumilli (changes requested):
@sbchaos any reason why tests were not included for these functions

sbchaos:
> @sbchaos any reason why tests were not included for these functions

Hey, the template functions are taken from the library we were using. Unfortunately the library does not have any test.
I was trying to add some test from my side. Will probably open another pr or with other changes.

PR #744, opened 2023-02-23
feat: change job version's default

arinda-arif:
Change the job spec version's default from 1 to 2. Any new job will use version 2 as default.


core/job/handler/v1beta1/job.go
```diff
@@ -240,7 +240,7 @@ func (jh *JobHandler) ListJobSpecification(ctx context.Context, req *pb.ListJobS
 
 func (*JobHandler) GetWindow(_ context.Context, req *pb.GetWindowRequest) (*pb.GetWindowResponse, error) {
 	// TODO: the default version to be deprecated & made mandatory in future releases
-	version := 1
```
sravankorumilli, line 243: lets avoid this, as GTF will be impacted with this

PR #753, opened 2023-03-03
feat: add replay

arinda-arif:


core/scheduler/service/replay_manager.go
```diff
@@ -0,0 +1,97 @@
...
+	if err != nil {
+		m.l.Error("unable to get on going replay")
+	}
+
+	for _, replay := range onGoingReplays {
+		runningTime := m.Now().Sub(replay.CreatedAt())
+		if runningTime < m.config.ReplayTimeout {
+			continue
+		}
+		message := fmt.Sprintf("replay is timing out. %s", replay.Message())
```
sravankorumilli, line 91: is there a scenario where replay message is updated twice if there is it is better to prefix it. & we could rename the message as "replay timed out"

core/scheduler/service/replay_manager.go
```diff
@@ -0,0 +1,97 @@
...
+		if err := m.replayRepository.UpdateReplayStatus(ctx, replay.ID(), scheduler.ReplayStateFailed, message); err != nil {
+			m.l.Error("unable to mark replay %s as failed due to time out", replay.ID())
+		}
+		m.l.Info("replay %s is timing out. marked as failed.", replay.ID())
```
sravankorumilli, line 95: similar replay timed out.

core/scheduler/service/replay_service.go
```diff
@@ -0,0 +1,56 @@
...
+	validator ReplayValidator
+}
+
+func (r ReplayService) CreateReplay(ctx context.Context, tenant tenant.Tenant, jobName scheduler.JobName, config *scheduler.ReplayConfig) (replayID uuid.UUID, err error) {
+	subjectJob, err := r.jobRepo.GetJobDetails(ctx, tenant.ProjectName(), jobName)
+	if err != nil {
+		return uuid.Nil, fmt.Errorf("unable to get job details from DB for jobName: %s, project:%s,  error:%w ", jobName, tenant.ProjectName().String(), err)
+	}
+
+	jobCron, err := cron.ParseCronSchedule(subjectJob.Schedule.Interval)
```
sravankorumilli, line 40: in which scenario job is add but cron can't be parsed

sravankorumilli: if this is not expected just add explicitly this shouldn't happen in a realistic scenario

internal/store/postgres/migrations/000050_replace_replay.up.sql
```diff
@@ -0,0 +1,29 @@
...
+    status  VARCHAR(30) NOT NULL,
+    message TEXT,
+
+    created_at TIMESTAMP WITH TIME ZONE NOT NULL,
+    updated_at TIMESTAMP WITH TIME ZONE NOT NULL
+);
+
+CREATE INDEX IF NOT EXISTS replay_request_project_name_idx on replay_request(project_name);
+CREATE INDEX IF NOT EXISTS replay_request_job_name_idx on replay_request(job_name);
+CREATE INDEX IF NOT EXISTS replay_request_start_time_idx on replay_request(start_time);
```
sravankorumilli, line 28: we don't need index on start & end_time just job_name should be sufficient.

sravankorumilli: as the table is very small we don't need to look into optimizing by creating many indexes

sravankorumilli (changes requested):

config/config_server.go
```diff
@@ -53,3 +56,10 @@ type ResourceManagerConfigOptimus struct {
 type PluginConfig struct {
 	Artifacts []string `mapstructure:"artifacts"`
 }
+
+// TODO: add worker interval
+type ReplayConfig struct {
+	WorkerTimeout time.Duration `mapstructure:"worker_timeout" default:"10m"`
```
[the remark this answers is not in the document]
arinda-arif: agreed to not configure it.

config/config_server.go
```diff
@@ -53,3 +56,10 @@ type ResourceManagerConfigOptimus struct {
...
+	ReplayTimeout time.Duration `mapstructure:"replay_timeout" default:"3h"`
+	LoopTimeout   time.Duration `mapstructure:"replay_timeout" default:"5m"`
```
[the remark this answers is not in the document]
arinda-arif: agreed to not configure it.

client/cmd/replay/create.go
```diff
@@ -0,0 +1,154 @@
...
+
+	"github.com/odpf/optimus/client/cmd/internal"
+	"github.com/odpf/optimus/client/cmd/internal/connectivity"
+	"github.com/odpf/optimus/client/cmd/internal/logger"
+	"github.com/odpf/optimus/config"
+	pb "github.com/odpf/optimus/protos/odpf/optimus/core/v1beta1"
+)
+
+const (
+	replayTimeout = time.Minute * 60
```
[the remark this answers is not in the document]
arinda-arif: reduce the number, 1 minute should be enough

core/scheduler/service/replay_manager.go
arinda-arif: message should not be updated twice.

arinda-arif: will change to `timed out`.

core/scheduler/service/replay_service.go
arinda-arif: in the error message, should mention this is not the expected scenario.

PR #757, opened 2023-03-13
fix: avoid order by and limit to make query performant

Mryashbhardwaj:
this is a to make the get queries performant

### DIFF
--- a/internal/store/postgres/scheduler/job_operator_repository.go
+++ b/internal/store/postgres/scheduler/job_operator_repository.go
@@ -19,7 +19,7 @@ const (
 	hookRunTableName   = "hook_run"
 
 	jobOperatorColumnsToStore = `name, job_run_id, status, start_time, end_time`
-	jobOperatorColumns        = `id, ` + jobOperatorColumnsToStore
+	jobOperatorColumns        = `id, ` + jobOperatorColumnsToStore + `, created_at`
 )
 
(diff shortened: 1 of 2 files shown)


internal/store/postgres/scheduler/job_operator_repository.go
```diff
@@ -74,22 +74,38 @@ func (o operatorRun) toOperatorRun() (*scheduler.OperatorRun, error) {
...
-	var opRun operatorRun
 	operatorTableName, err := operatorTypeToTableName(operatorType)
 	if err != nil {
 		return nil, err
 	}
-	getJobRunByID := "SELECT " + jobOperatorColumns + " FROM " + operatorTableName + " j where job_run_id = $1 and name = $2 order by created_at desc limit 1"
-	err = o.db.QueryRow(ctx, getJobRunByID, jobRunID, name).
-		Scan(&opRun.ID, &opRun.Name, &opRun.JobRunID, &opRun.Status, &opRun.StartTime, &opRun.EndTime)
-
+	getJobRunByID := "SELECT " + jobOperatorColumns + " FROM " + operatorTableName + " j where job_run_id = $1 and name = $2"
```
sbchaos, line 81: Have we done a performance comparison for this, is the query returning multiple records faster than the one with order by in the query.

PR #765, opened 2023-03-24
feat: replay config overriding

deryrahman:


internal/store/postgres/scheduler/replay_repository.go
```diff
@@ -252,6 +254,30 @@ func (r ReplayRepository) UpdateReplay(ctx context.Context, id uuid.UUID, replay
 	return r.updateReplayRuns(ctx, id, runs)
 }
 
+func (r ReplayRepository) GetReplayTaskConfigByScheduledAt(ctx context.Context, scheduledAt time.Time) (map[string]string, error) {
```
arinda-arif, line 257: Any reason why we are not also filtering with the job name?

internal/store/postgres/scheduler/replay_repository_test.go
```diff
@@ -19,6 +19,7 @@ func TestPostgresSchedulerRepository(t *testing.T) {
 	tnnt, _ := tenant.NewTenant("test-proj", "test-ns")
 	endTime := time.Now()
 	startTime := endTime.Add(-48 * time.Hour)
+	replayJobConfig := map[string]string{"EXECUTION_PROJECT": "example_project"}
```
arinda-arif, line 22: let's add a single test that inserts replay without any job config specified

arinda-arif (changes requested):

internal/store/postgres/scheduler/replay_repository.go
deryrahman: It should use job name as well, will create a patch on it

